PR #105, opened 2023-11-21
Populate more JavacOptionsResult info

Arthurm1:
Adds `classpath` and `classesDir` to `JavacOptionsItem`

### DIFF
--- a/model/src/main/java/com/microsoft/java/bs/gradle/model/GradleSourceSet.java
+++ b/model/src/main/java/com/microsoft/java/bs/gradle/model/GradleSourceSet.java
@@ -70,6 +70,11 @@ public interface GradleSourceSet extends Serializable {
   public File getResourceOutputDir();
 
   /**
+   * The compile classpath for this source set.
+   */
+  public List<File> getCompileClasspath();
+
+  /**
    * JDK home file location.
    */
   public File getJavaHome();
--- a/model/src/main/java/com/microsoft/java/bs/gradle/model/impl/DefaultGradleSourceSet.java
+++ b/model/src/main/java/com/microsoft/java/bs/gradle/model/impl/DefaultGradleSourceSet.java
@@ -37,6 +37,8 @@ public class DefaultGradleSourceSet implements GradleSourceSet {
 
   private File sourceOutputDir;
 
+  private List<File> compileClasspath;
+
   private Set<File> resourceDirs;
 
   private File resourceOutputDir;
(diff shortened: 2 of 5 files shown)


Arthurm1:

jdneo:
Hi @Arthurm1,

Thank you for your contribution! And apologize for the late reply (It's the release week from my side).

I will review all your PRs ASAP!

jdneo (commented):
Thank you @Arthurm1,

It looks good to me overall, just some Checkstyle violations. We can merge this PR after they are fixes.

Arthurm1:
@jdneo I've fixed the checkstyle issues - I hadn't realised checkstyle was being used.

It would be good if checkstyle ran automatically on every PR that was submitted.

jdneo (approved):
LGTM

jdneo:
@Arthurm1 Thank you for the contribution.

I'm a little bit surprised that you created so many PRs, just asking out of curiosity, are you trying to integrate the Gradle build server somewhere?

Arthurm1:
@jdneo lol - yes.

I'm trying to integrate it with [Metals](https://github.com/scalameta/metals) which is a Scala LSP but also does Java.

Currently Gradle is supported in Metals by exporting the config to [Bloop](https://github.com/scalacenter/bloop) (a Java/Scala build server) using [Gradle Bloop plugin](https://github.com/scalacenter/gradle-bloop) which I've contributed to - hence the weird test cases I'm adding here which have sprung up over the years.

One of the downsides of using Gradle + Bloop is that generated sources aren't really handled well and there's no good route for this as projects often define their own custom tasks for this which aren't easy to extract to pass to Bloop.

If I have time, I'd like to add **run** and **test** to this project and to handle the **scala plugin** which I think should all be possible.  I see that the Gradle team are also creating a Problems API so in the future compiler diagnostics can be handled and sent back to the BSP client.

One thing I have no idea about is handling of debugging.

Do you have a roadmap on your plans for this project?

jdneo:
It's so exciting to hear that! More adoption will definitely improve the product!

I haven't a concrete ETA about the next big parts to deliver. But the things you mentioned are on our plan exactly.

> I'd like to add run and test to this project

We are also considering delegating the test execution to the Gradle. Because for a Gradle project, test source sets can have their own classpath, while that concept does not apply to our project model in the Java Language server. So running Gradle tests is not well supported in VS Code Java

> handle the scala plugin

That's great. Maybe we can think of some kind of contribution points allowing 'injecting' plugins to handle different kind of build targets. Besides Java & scala, we can also have things like Android, etc...

I'll try my best to review the opened PRs recently. Thank you again for your contributions. Let's keep in touch :)

PR #106, opened 2023-11-22
Populate buildTarget/displayName

Arthurm1:
This PR populates display name in `BuildTarget` which is currently null.

Creating unique names is a bit fiddly as BSP considers each source set a separate target and unique whereas Gradle considers each project unique.  I've added a test for this.

### DIFF
--- a/plugin/src/main/java/com/microsoft/java/bs/gradle/plugin/SourceSetsModelBuilder.java
+++ b/plugin/src/main/java/com/microsoft/java/bs/gradle/plugin/SourceSetsModelBuilder.java
@@ -113,7 +113,19 @@ public class SourceSetsModelBuilder implements ToolingModelBuilder {
         gradleSourceSet.setCompilerArgs(getCompilerArgs(project, sourceSet));
         gradleSourceSets.add(gradleSourceSet);
       });
+    }
 
+    // run through twice to cope with any name clashes
+    for (DefaultGradleSourceSet sourceSet : sourceSetMap.values()) {
+      if ("main".equals(sourceSet.getSourceSetName())) {
+        sourceSet.setDisplayName(stripPathPrefix(sourceSet.getProjectPath()));
+      }
+    }
+    for (DefaultGradleSourceSet sourceSet : sourceSetMap.values()) {
+      if (!"main".equals(sourceSet.getSourceSetName())) {
+        String uniqueName = createUniqueDisplayName(sourceSet, sourceSetMap);
+        sourceSet.setDisplayName(uniqueName);
+      }
     }
 
     for (Project project : allProject) {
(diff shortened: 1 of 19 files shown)


plugin/src/main/java/com/microsoft/java/bs/gradle/plugin/SourceSetsModelBuilder.java
```diff
@@ -139,6 +150,32 @@ public Object buildAll(String modelName, Project rootProject) {
...
+    Map<SourceSet, DefaultGradleSourceSet> sourceSetMap) {
+
+    String fullName = stripPathPrefix(sourceSet.getProjectPath()) + "-" + sourceSet.getSourceSetName();
+    // has the suffix caused a clash - apply a numbered suffix
+    String usedName = fullName;
+    int i = 2;
+    boolean found = true;
+    while (found) {
+      String nameToTest = usedName;
+      if (!sourceSetMap.values().stream().anyMatch(ss -> nameToTest.equals(ss.getDisplayName())))
```
jdneo: Let's use `noneMatch` instead of negated `anyMatch`

jdneo (commented):
There are some Checkstyle vioaltions. (Sorry the CI won't automatically run when the PRs are from outside contributors)

If you don't mind, you can install the [Checkstyle](https://marketplace.visualstudio.com/items?itemName=shengchen.vscode-checkstyle) extension. The check should be enabled automatically since I've configured the settings in the `settings.json`. It will highlight the violations in the editor on the fly during editing.

Or running the Grdle task `check` to find the Checkstyle violations.

------------

One more question about the background. Would you mind sharing in which case will the `displayName` be used at the client?

Arthurm1:
Display name in Metals is shown as a user-friendly way to navigate build targets.
There is a project explorer which, after this PR, displays as...
image

There is also a "Doctor" page to show what Metals thinks about the project setup...
image

And finally you can get info on each build target...
image


I can also see a case where someone has written a command line BSP client and a user might want to manage the build through it.  In that case they could write something simple like `bsp compile model model-test`.  Without these display names, they'd have to type the entire build target uri which is long and not user-friendly.

plugin/src/main/java/com/microsoft/java/bs/gradle/plugin/SourceSetsModelBuilder.java
Arthurm1: Done

Arthurm1:
I've made the changes - should be good to merge now

jdneo (approved):
LGTM. Thank you @Arthurm1

plugin/src/main/java/com/microsoft/java/bs/gradle/plugin/SourceSetsModelBuilder.java
```diff
@@ -113,7 +113,19 @@ public Object buildAll(String modelName, Project rootProject) {
...
 
+    // run through twice to cope with any name clashes
+    for (DefaultGradleSourceSet sourceSet : sourceSetMap.values()) {
+      if ("main".equals(sourceSet.getSourceSetName())) {
+        sourceSet.setDisplayName(stripPathPrefix(sourceSet.getProjectPath()));
+      }
+    }
+    for (DefaultGradleSourceSet sourceSet : sourceSetMap.values()) {
+      if (!"main".equals(sourceSet.getSourceSetName())) {
+        String uniqueName = createUniqueDisplayName(sourceSet, sourceSetMap);
```
jdneo, line 126: @Arthurm1 I just had this new question in my mind when I was trying to refactor the code.

Why not just directly append the sourceset name to the project path string? for example:

`:a:b [main]` & `:a:b [test]`.

By doing this, looks like we can remove the logic to check duplicated names.

Arthurm1: You can.  But I was trying to get the name as simple as possible for the user.

If you look at Gradle tasks for the `model` project you have tasks like...
main source set: `classes, compileJava`, test source set: `testClasses, compileTestJava`

So Gradle tries to just ignore the `main` in the `main` source set.

Also people use text based editors e.g. vim, emacs, helix etc. and CLI for BSP so it's good to have these names as simple and short as possible if you ever have to type them.
I'd like to avoid special characters (space, braces) for this same reason.  But that might be more an issue for the client to deal with than the server as there's nothing in the BSP spec about displayName limitations.

I guess it's personal preference.
This...
image
Or this...
image

I find the first one easier on the eye.

jdneo: I see. Thank you for the explanation. Makes sense since Gradle has similar concept for the tasks.

> I guess it's personal preference.

Yes. Because I feel that when there is a conflict, things like `foo-test` & `foo-test2`might be a little bit strange. It's just my personal opinion, and I don't have a strong opinion which one is better.

PR #107, opened 2023-11-23
handle different project dependency configs

Arthurm1:
The current way of populating project dependencies doesn't always return all the dependencies.  This is mainly because Gradle doesn't really expose the concept of a sourceset in one project depending on a sourceset in another project.

I've added a number of tests to show how various configs have got round this limitation over the years.

I don't think the current Gradle API allows a way to simply query the project dependencies that would satisfy all these new tests.

So the implementation here queries a project's classpath and then matches those entries to the output paths (or archive jars) of other projects' sourcesets.


Arthurm1:
Fixed checkstyle and rebased

model/src/main/java/com/microsoft/java/bs/gradle/model/GradleProjectDependency.java
```diff
@@ -10,4 +10,6 @@
  */
 public interface GradleProjectDependency extends Serializable {
```
jdneo: I guess we need to rename it to something like `GradleSourceSetDependency`?

plugin/src/main/java/com/microsoft/java/bs/gradle/plugin/SourceSetsModelBuilder.java
```diff
@@ -137,13 +193,24 @@ public Object buildAll(String modelName, Project rootProject) {
...
 
     return new DefaultGradleSourceSets(gradleSourceSets);
   }
 
+  private Set<Object> getArchiveSourcePaths(CopySpec copySpec) {
+    Set<Object> sourcePaths = new HashSet<>();
+    if (copySpec instanceof DefaultCopySpec) {
+      DefaultCopySpec defaultCopySpec = (DefaultCopySpec) copySpec;
+      sourcePaths.addAll(defaultCopySpec.getSourcePaths());
+      for (CopySpec child : defaultCopySpec.getChildren()) {
```
jdneo: I tested on an old gradle (4.9) and got:

```
Caused by: java.lang.NoSuchMethodError: org.gradle.api.internal.file.copy.DefaultCopySpec.getChildren()Ljava/util/Collection;
```

Arthurm1: I've pushed an update (minus a lot of the java test files as they're not needed, just the build setup is needed).
This should solve the 4.9 issue but I'm struggling to test this on early gradle versions - could you test it yourself again please?
Also rebased

testProjects/project-dependency-configuration/a/build.gradle
```diff
@@ -0,0 +1,8 @@
+plugins {
+	id 'java'
+	id 'java-gradle-plugin'
```
jdneo: A lot of testing projects contains the `java-gradle-plugin`, which looks like can be removed.

model/src/main/java/com/microsoft/java/bs/gradle/model/GradleProjectDependency.java
jdneo: The reason that I found the name may need to be change is because, we have some tests like `testGetGradleDependenciesWithTestToMain()`. This test tests that the main source set depends on test source set. But, both two source set belongs to the same project.

This is different from the definition of `project dependency` from Gradle document. I'm not sure if it's still proper to call it a project dependency.

Arthurm1: I can change the name.  How about `BuildTargetDependency`?  All the `XXXSourceSet` named classes could be changed to `XXXBuildTarget`?

testProjects/project-dependency-configuration/a/build.gradle
Arthurm1: You're right.  I copied the build defs from `junit5-jupiter-starter-gradle` but they're not required.  I've removed them and pushed the changes

model/src/main/java/com/microsoft/java/bs/gradle/model/GradleProjectDependency.java
jdneo: `BuildTargetDependency` looks good. And we can also update the javadoc explaining that it could be a project dependency or source set dependency.

plugin/src/main/java/com/microsoft/java/bs/gradle/plugin/SourceSetsModelBuilder.java
```diff
@@ -207,6 +266,36 @@ private String createUniqueDisplayName(DefaultGradleSourceSet sourceSet,
     return usedName;
   }
 
+  private Set<Object> getArchiveSourcePaths(CopySpec copySpec) {
+    Set<Object> sourcePaths = new HashSet<>();
+    if (copySpec instanceof DefaultCopySpec) {
+      DefaultCopySpec defaultCopySpec = (DefaultCopySpec) copySpec;
+      sourcePaths.addAll(defaultCopySpec.getSourcePaths());
+      // DefaultCopySpec#getChildren changed from Iterable to Collection
+      if (GradleVersion.current().compareTo(GradleVersion.version("6.2")) >= 0) {
```
jdneo, line 275: What about

```java

Object children = defaultCopySpec.getChildren();
if (children instanceof Collection) {
    ...
} else if (child instanceof Iterable) {
    ...
}
```

model/src/main/java/com/microsoft/java/bs/gradle/model/GradleProjectDependency.java
Arthurm1: I've changed the `GradleProjectDependency` to `BuildTargetDependency`.

plugin/src/main/java/com/microsoft/java/bs/gradle/plugin/SourceSetsModelBuilder.java
Arthurm1: That would only work if we compiled the project under an earlier Gradle API version which used `Iterable`.  Which I don't think is a good idea as we'd be stuck on that version and we'd have to use reflection for other newer methods.
With the current version, the `getChildren` method returns a `Collection` and this is set in the `SourceSetsModelBuilder` class file.  That means that when it's run under an earlier version it will fail to find this method as the method in the earlier version returns `Iterable`.  Hence the use of reflection.
If you `javap` the `SourceSetsModelBuilder` class you will see what I mean.  Note the expectation of `Collection`...
```
   #262 = Methodref          #238.#263    // org/gradle/api/internal/file/copy/DefaultCopySpec.getChildren:()Ljava/util/Collection;
   #263 = NameAndType        #264:#104    // getChildren:()Ljava/util/Collection;
```

Arthurm1: Do you have an idea of the oldest version of Gradle you want to support?  The other option is making 6.2 the minimum version but I don't think that's a good idea.  From my experience, a lot of people use older versions of Gradle, partly because they are using plugins that are discontinued and/or don't support later versions of Gradle.

jdneo: > That means that when it's run under an earlier version it will fail to find this method as the method in the earlier version returns Iterable.

You are right. Thank you for the explanation!

> Do you have an idea of the oldest version of Gradle you want to support?

No idea for that. I was just trying the best to support as much as I can.

> From my experience, a lot of people use older versions of Gradle

True, data shows that lots of people are using Gradle < 4.0, which is a quite old version.

Arthurm1:
Fixed checkstyle issue

jdneo:
Looks good to me and able to be merged. Meanwhile, there is a PR at the client side needs to be merged based on this change.

I have a question that may not be related with this work:

What about composite build in Gradle: https://docs.gradle.org/current/samples/sample_composite_builds_basics.html

How does bloop handle it?

jdneo (approved):

Arthurm1:
I'm not sure.  I guess I could add a test and see what happens

jdneo:
I'll merge this once https://github.com/microsoft/vscode-gradle/pull/1464 is merged. It's holiday season now so please expect the delay.

PR #108, opened 2023-11-23
fix: Populate test tag correctly for all sourcesets

Arthurm1:
Currently the BSP `BuildTargetTag` tag is populated as `Test` only when the source set name is `test`.

All other build targets won't have this tag, which means other test configurations (e.g. integration tests) won't have the tag.

Now the code checks to see if there is a `Test` task associated with the source set and populates the tag from that information.


Arthurm1:
rebased

jdneo (commented):
Changes looks good. Need to rebase before merge.

Arthurm1:
Rebased

jdneo (approved):

jdneo:
Thank you @Arthurm1

PR #109, opened 2023-11-23
Populate java compiler args

Arthurm1:
Populate the missing javac options - before it was the only the additional used specified options that appeared.

### DIFF
--- a/plugin/src/main/java/com/microsoft/java/bs/gradle/plugin/SourceSetsModelBuilder.java
+++ b/plugin/src/main/java/com/microsoft/java/bs/gradle/plugin/SourceSetsModelBuilder.java
@@ -13,15 +13,21 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import org.gradle.api.Project;
 import org.gradle.api.file.Directory;
 import org.gradle.api.file.SourceDirectorySet;
+import org.gradle.api.internal.tasks.compile.DefaultJavaCompileSpec;
+import org.gradle.api.internal.tasks.compile.JavaCompilerArgumentsBuilder;
 import org.gradle.api.plugins.JavaPluginExtension;
 import org.gradle.api.tasks.SourceSet;
 import org.gradle.api.tasks.SourceSetContainer;
@@ -108,9 +114,11 @@ public class SourceSetsModelBuilder implements ToolingModelBuilder {
         gradleSourceSet.setJavaHome(defaultJavaHome);
         gradleSourceSet.setJavaVersion(javaVersion);
         gradleSourceSet.setGradleVersion(gradleVersion);
-        gradleSourceSet.setSourceCompatibility(getSourceCompatibility(project, sourceSet));
-        gradleSourceSet.setTargetCompatibility(getTargetCompatibility(project, sourceSet));
         gradleSourceSet.setCompilerArgs(getCompilerArgs(project, sourceSet));
+        gradleSourceSet.setSourceCompatibility(
+            getSourceCompatibility(gradleSourceSet.getCompilerArgs()));
+        gradleSourceSet.setTargetCompatibility(
+            getTargetCompatibility(gradleSourceSet.getCompilerArgs()));
         gradleSourceSets.add(gradleSourceSet);
       });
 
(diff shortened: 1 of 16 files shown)


nickzhums:
@Arthurm1 Thank you so much for the contribution! Since there is a lot of PRs pending, it's gonna take some time for review, please allow us some time :) again, thanks so much!

nickzhums:
@Arthurm1 meanwhile, feel free us to let us know if you have any feedback using the product, anything is useful to us, thanks

jdneo (approved):
Except for the Checkstyle violations, this change looks good to me.

With this PR merged, looks like we can get rid of

https://github.com/microsoft/build-server-for-gradle/blob/develop/model/src/main/java/com/microsoft/java/bs/gradle/model/GradleSourceSet.java#L92-L100

and recover the source and target compatibility from the compiler args. I'll do that later.

Arthurm1:
Fixed checkstyle issues

jdneo:
Since it's using internal apis, there might be chances to meet `NoSuchClass`/`NoSuchMethod` error.

Do we have public utils that can be used? If not, considering adding a try-catch block and use the original implementation as a fallback.

Arthurm1:
I've wrapped the code in try-catch as there are no public classes that create the full args list.  The other option is to duplicate the `JavaCompilerArgumentsBuilder` class but there's a lot of code in there and it's subject to change so it would have to be maintained.

I've added more tests as I wasn't handling the `--release` flag properly.

I've added code to cope with Groovy's use of `GStringImpl` when Strings are built using String interpolation e.g. `"${variable}"`.  This causes `IllegalArgumentException` when returning data through the tooling API so these have to be manually converted to Strings.  Yuck

plugin/src/main/java/com/microsoft/java/bs/gradle/plugin/SourceSetsModelBuilder.java
```diff
@@ -314,7 +318,49 @@ private String getTargetCompatibility(Project project, SourceSet sourceSet) {
...
     if (javaCompile != null) {
-      return javaCompile.getOptions().getCompilerArgs();
+      CompileOptions options = javaCompile.getOptions();
+
+      try {
+        DefaultJavaCompileSpec specs = new DefaultJavaCompileSpec();
+        specs.setCompileOptions(options);
+
+        if (GradleVersion.current().compareTo(GradleVersion.version("6.6")) >= 0) {
+          if (javaCompile.getOptions().getRelease().isPresent()) {
```
jdneo: Since the api document says: `If set, it will take precedences over the getSourceCompatibility() and getTargetCompatibility() settings.`

what about extract the spec setup logics into a method, and make an early return when `release` is available? Then there is no need to check `specs.getRelease() == null` in the below two if blocks.

plugin/src/main/java/com/microsoft/java/bs/gradle/plugin/SourceSetsModelBuilder.java
```diff
@@ -314,7 +318,49 @@ private String getTargetCompatibility(Project project, SourceSet sourceSet) {
...
+                .includeLauncherOptions(false);
+        return builder.build();
+      } catch (Exception e) {
+        // DefaultJavaCompileSpec and JavaCompilerArgumentsBuilder are internal so may not exist.
+        // Fallback to returning just the compiler arguments the build has specified.
+        // This will miss a lot of arguments derived from the CompileOptions e.g. sourceCompatibilty
+        // Arguments must be cast and converted to String because Groovy can use GStringImpl
+        // which then throws IllegalArgumentException when passed back over the tooling connection.
+        List<Object> compilerArgs = new LinkedList<>(options.getCompilerArgs());
+        return compilerArgs
```
jdneo, line 390: The return of `getCompilerArgs()` is already a `List<String>`, should be no need to call `toString()` for all the elements

plugin/src/main/java/com/microsoft/java/bs/gradle/plugin/SourceSetsModelBuilder.java
```diff
@@ -314,7 +318,49 @@ private String getTargetCompatibility(Project project, SourceSet sourceSet) {
...
+        }
+
+        JavaCompilerArgumentsBuilder builder = new JavaCompilerArgumentsBuilder(specs)
+                .includeMainOptions(true)
+                .includeClasspath(false)
+                .includeSourceFiles(false)
+                .includeLauncherOptions(false);
+        return builder.build();
+      } catch (Exception e) {
+        // DefaultJavaCompileSpec and JavaCompilerArgumentsBuilder are internal so may not exist.
```
jdneo, line 384: A thing I want to have some discussion a little bit.

I was intended to remove `getSourceCompatibility()` and `getTargetCompatibility()` with the assumption that the `release/source/target` args are always available with this new approach.

Since it's not for old version Gradle, I cannot simply remove it now. The source and target level are used by the java language server to pick the right project jdk.

Do you think it makes sense to manually add `release/source/target` when internal util not exist?

Arthurm1: You mean if none of the paths here are triggered and so none of `-source`, `-target`, `--release` make it into the compiler args?  I guess it should then pick this up from the jvm that gradle is running under.  I suppose we should also set the source/target in the DefaultGradleSourceset so jvmBuildTargetEx picks it up.  I should probably add a test for toolchains as well, in case they set the source/target/release in a different manner.

Arthurm1:
Hopefully I've addressed all your points.

The source and target compatibility are now both set from the compiler args.  Before, if a build had set the source/target directly in the args, then the compatibility settings would be wrong.
If nothing is set in the project's build file then Gradle automatically adds the current JVM version to the compiler args.  So these should always be set in DefaultGradleSourceSet and hence in JvmBuildTargetEx.

I've added to the tests to check for both these issues.

plugin/src/main/java/com/microsoft/java/bs/gradle/plugin/SourceSetsModelBuilder.java
jdneo: Directly `return new LinkedList<>(options.getCompilerArgs());`?

jdneo: > Arguments must be cast and converted to String because Groovy can use GStringImpl, which then throws IllegalArgumentException when passed back over the tooling connection.

Ok, I see.

plugin/src/main/java/com/microsoft/java/bs/gradle/plugin/SourceSetsModelBuilder.java
jdneo: I was trying to avoid extending the spec as much as I can. The `JvmBuildTargetEx` extends the spec and it looks like at least the fields `sourceCompatibility` and `targetCompatibility` are not necessary, because they can be covered by the javac options.

Then, if `JavaCompilerArgumentsBuilder` is available, the `sourceCompatibility` and `targetCompatibility` are available as well. On the other hand, we can manually append `--source/--target` or `--release` to the returned string list. Then it should be ok to remove `sourceCompatibility` and `targetCompatibility` from `JvmBuildTargetEx`. The client can check the javac options and get those information.

It's not necessary to cover that in this PR. We can do that when addressing [#112](https://github.com/microsoft/build-server-for-gradle/issues/112). Meanwhile, please let me know if the above assumption is wrong from you point of view.

jdneo (approved):
Looks good to me.

> I should probably add a test for toolchains as well, in case they set the source/target/release in a different manner

Would you like to add a toolchain test in this PR? I'm both ok if merging the PR as it is.

plugin/src/main/java/com/microsoft/java/bs/gradle/plugin/SourceSetsModelBuilder.java
Arthurm1: I think that's fine and if the BSP client needs the source/target then it can parse the compiler args.

Arthurm1:
@jdneo I've added a toolchain test.  No other changes.  Hopefully good to merge now

jdneo (approved):

jdneo:
Thank you @Arthurm1!

PR #110, opened 2023-11-25
implement buildTargetCleanCache

Arthurm1:
implements clean

### DIFF
--- a/server/src/main/java/com/microsoft/java/bs/core/internal/services/BuildTargetService.java
+++ b/server/src/main/java/com/microsoft/java/bs/core/internal/services/BuildTargetService.java
@@ -15,6 +15,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
+import java.util.function.Function;
 import java.util.logging.Level;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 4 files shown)


server/src/main/java/com/microsoft/java/bs/core/internal/services/BuildTargetService.java
```diff
@@ -347,6 +369,25 @@ private String getBuildTaskName(BuildTargetIdentifier btId) {
...
+      throw new IllegalArgumentException("The build target does not exist: " + btId.getUri());
+    }
+    GradleSourceSet sourceSet = gradleBuildTarget.getSourceSet();
+    String classesTaskName = "clean";
+
+    String modulePath = sourceSet.getProjectPath();
+    if (modulePath == null || modulePath.equals(":")) {
+      return classesTaskName;
+    }
+    return modulePath + ":" + classesTaskName;
```
jdneo, line 385: How about extracting a method something like `String getTaskName(String modulePath, String taskName) {...}`.

Then it can be re-used in both `getCleanTaskName()` and `getBuildTaskName()`

jdneo (commented):
Thank you @Arthurm1 

Just one small piece of comment.

server/src/main/java/com/microsoft/java/bs/core/internal/services/BuildTargetService.java
Arthurm1: I've extracted the common code to a `runTasks` method

jdneo (approved):
LGTM.

jdneo:
Thank you @Arthurm1 for the contribution! I've appended a commit updating the readme for this new capability.

PR #113, opened 2023-12-05
Add Scala support

Arthurm1:
Implements `buildTarget/scalacOptions`.  Also returns `ScalaBuildTarget` instead of `JvmBuildTarget` if Scala has been configured for that target and passes back `scala` as a supported language.

I haven't separated out the implementation per language.  I think a refactor would be better done in a single PR (not one where a new language is added).  I'm not sure it's currently worth refactoring.  Maybe when more languages are added?

Creating the scala compiler args using `ZincScalaCompilerArgumentsGenerator` is a bit verbose because of the class it requires to extract the setup from.  It only actually needs 2 methods to be implemented.  Another option would be to copy the code directly from `ZincScalaCompilerArgumentsGenerator` into `SourceSetsModelBuilder`.  I thought this would be easier to maintain if the original is changed.


jdneo:
> I'm not sure it's currently worth refactoring. Maybe when more languages are added?

To me, I would like to have a refactoring at current stage. Because:
- The current `GradleSourceSet.java` has become a large object with a lot of fields.
- If we defer the refactoring to the third language comes, things will be even more complex than now.

I'm sorry that this is a debt that I missed when implementing.

About the refactoring, I'm thinking that

- We can have an extensibility that allows third-party plugin to provide their language build targets.
- For the current embedded plugin, it can have different model builders taking care of their own language models.

**We can do the second refactoring first to unblock the scala support**.

The rough idea in my mind is:
- the plugin will register the model builder according to the `languageIds` of the client capability.
- The Gradle source set holds the common shared data.
- The Gradle source set has an extension that holds language specific data. (like Scala version)

WDYT?

Arthurm1:
> the plugin will register the model builder according to the languageIds of the client capability.
> The Gradle source set holds the common shared data.
> The Gradle source set has an extension that holds language specific data. (like Scala version)

It sounds OK.  I'm happy to code around your design, I'm just don't know whether that's the right direction as I haven't seen enough different language setups in Gradle to make a decision on splitting the design across languages.

One big advantage here is that I doubt anyone will be using this project as a library (it doesn't make sense to me to do that) so I think you can change the design/API as and when you like without affecting downstream projects.

Just some thoughts, not blockers...
1) A single Build Target can have multiple language types so whatever these extensions are, the common gradle source set would need to be mapped to extensions 1->many, not 1->1.  E.g scala can mix and match java and scala code within the same gradle project configuration.  It would need both Scala extension info and Java extension info.  So I wouldn't like to see the extension classes inherit from the common info class.

2) We might want to build the common set regardless of whether the languages are supported by the BSP client.  One example I can think of is that a multi-project might have mostly Java subprojects but have the test projects written in Kotlin.  A Java only client could still run these tests without knowing anything about the Kotlin language and debug into the Java code.  The BSP compile/run/test/debug requests are all language agnostic.

3) Would this work nicely for Android? I've written some code to extract Gradle Android info before [here](https://github.com/scalacenter/gradle-bloop/blob/c21a6bd7c04e6a40634c7466cab4f99074b82712/src/main/scala/bloop/integrations/gradle/model/BloopConverter.scala#L74).  Annoyingly it ignores the Gradle sourceset structures and defines its own.  So the common shared data class might not actually include sourcesets (but would include dependencies/modules/names etc).
Would the Android extract be registered by language?  I'm not sure that makes sense as the extracted sourcesets are still Java/Scala/Kotlin.  Unless the new structure allows multiple ModelBuilders per language.
AFAIK - you have to register your ToolingModelBuilder by the class it produces so I guess you'd have basic wrapper classes e.g. AndroidCommon, AndroidJava, AndroidScala, AndroidKotlin which just wrap the extension classes written for those languages?

jdneo:
Thank you @Arthurm1 for your input.

Here is my plan: I'll first make sure those opened PR get merged recently. At the meantime, I'll find some time think thoroughly about how to do the refactoring. We can continue the discussion here.

jdneo:
> Annoyingly it ignores the Gradle sourceset structures and defines its own. So the common shared data class might not actually include sourcesets

I'm not very familiar with Android, which kind of property of the source set is not aligned when it comes to Android projects?

jdneo:
@Arthurm1 I now have a feeling that, maybe the BSP related types should be constructed at the plugin side, and the server is just simply get them from plugins and return them when different requests come.

Currently the server needs to know the definition of the source set and parse them to build target, etc... This makes trouble as you've mentioned, the concept of source set does not exist for Android, not to say other languages.

Arthurm1:
> I now have a feeling that, maybe the BSP related types should be constructed at the plugin side

Maybe - I don't think that's a bad idea - I'm just not sure yet.

Android does have the concept of sourcesets - it just has it's own hierarchy.
So instead of [`org.gradle.api.tasks.SourceSet`](https://github.com/gradle/gradle/blob/master/platforms/jvm/platform-jvm/src/main/java/org/gradle/api/tasks/SourceSet.java) it has [`com.android.build.api.dsl.AndroidSourceSet`](https://developer.android.com/reference/tools/gradle-api/8.3/com/android/build/api/dsl/AndroidSourceSet)

What I do think is that this data might have to be cached.  Have you tried the plugin on a big project? e.g. [Gradle itself](https://github.com/gradle/gradle).  It takes about a minute to respond to build/initialize (which I think should be instant and is slow because the source sets are queried in response to this request but not yet needed).  Then, after a compile, the source sets are all queried again.  It would be good to delay the building of the sourcesets until the client requests them and do they need to be built again after every compile?

jdneo:
About the perf issue, let's discuss at https://github.com/microsoft/build-server-for-gradle/issues/118.

> Android does have the concept of sourcesets - it just has it's own hierarchy.
So instead of [org.gradle.api.tasks.SourceSet](https://github.com/gradle/gradle/blob/master/platforms/jvm/platform-jvm/src/main/java/org/gradle/api/tasks/SourceSet.java) it has [com.android.build.api.dsl.AndroidSourceSet](https://developer.android.com/reference/tools/gradle-api/8.3/com/android/build/api/dsl/AndroidSourceSet)

Does that mean that for android project, we need a different way to fetch the information of the source set, but the properties of the source set are the same?

Arthurm1:
> Does that mean that for android project, we need a different way to fetch the information of the source set, but the properties of the source set are the same?

I think so.

jdneo:
@Arthurm1 I have got some ideas about the refactoring. I'll make a draft these days and let you review it.

plugin/src/main/java/com/microsoft/java/bs/gradle/plugin/SourceSetsModelBuilder.java
```diff
@@ -76,15 +83,24 @@ public Object buildAll(String modelName, Project rootProject) {
...
+
+        // java
+        JavaCompile javaCompile = getJavaCompileTask(project, sourceSet);
+        if (javaCompile != null) {
+          gradleSourceSet.setIsJava(true);
+          gradleSourceSet.setSourceCompatibility(getSourceCompatibility(javaCompile));
+          gradleSourceSet.setTargetCompatibility(getTargetCompatibility(javaCompile));
+          gradleSourceSet.setJavaCompilerArgs(getJavaCompilerArgs(javaCompile));
+          Set<File> generatedSrcDirs = new HashSet<>();
+          addAnnotationProcessingDir(javaCompile, generatedSrcDirs);
```
jdneo: Does that mean Scala does not need dealing with the annotation processing?

Since scala plugin is extended from java plugin, when it comes to a project mixed with java and scala, is it possible that using annotation processing for the java part?

Arthurm1: The project can be using annotation processing on the java part.  The `JavaCompile` task should then exist and the compiler args and annotation processing directory should be set here.
So the project can have both `isJava` and `isScala` set to true.  Then a `ScalaBuildTarget` is returned which contains the `JavaBuildTarget` information as well.

jdneo: Forget about it, I'm stupid. Since Scala plugin is extended from Java plugin, so JavaCompile task should be always available.

plugin/src/main/java/com/microsoft/java/bs/gradle/plugin/SourceSetsModelBuilder.java
```diff
@@ -127,89 +140,87 @@ public Object buildAll(String modelName, Project rootProject) {
...
+
+        // scala.  Requires module dependencies to resolve version info.
+        ScalaCompile scalaCompile = getScalaCompileTask(project, sourceSet);
+        if (scalaCompile != null) {
+          GradleModuleDependency scalaLibraryDependency =
+              getScalaLibraryDependency(gradleSourceSet.getModuleDependencies());
+          if (scalaLibraryDependency != null) {
+            gradleSourceSet.setIsScala(true);
+            gradleSourceSet.setScalaOrganization(getScalaOrganization(scalaLibraryDependency));
+            gradleSourceSet.setScalaVersion(getScalaVersion(scalaLibraryDependency));
```
jdneo: Hi @Arthurm1, just a question, no action needs to take.

Today when I was reading the spec of BSP, I found following statements:

```
a single configuration in a single project that contains both Java and Scala sources maps to one BuildTarget.
a project with crossScalaVersions 2.11 and 2.12 containing main and test configuration in each will have 4 build targets.
a Scala 2.11 and 2.12 cross-built project for Scala.js and the JVM with main and test configurations will have 8 build targets.
```

I'm pretty sure that current implementation can apply the first one. What about the second and third?

Arthurm1: This is really for other build tools that support BSP e.g. SBT, Mill, Bazel.  It's not possible to do cross versioning in Gradle with the built in Scala plugin.  Unless there's a 3rd party plugin somewhere around that I'm not aware of.

jdneo: Ah, I see. Thank you!

jdneo:
@Arthurm1, Sorry bro for the long time hanging. Would you like to continue the work based on the latest code base?

Arthurm1:
@jdneo In the first commit, I've refactored for the new multi-language setup.  The only thing of note is I've moved `compileClasspath` back into `GradleSourceSet` as it seems to be common across all languages.

In the second commit, I've refactored the extension conversion.  I've managed to make it so the conversion to a `JavaExtension` or `ScalaExtension` is done once in the `DefaultGradleSourceSet` constructor.  It still uses reflection sadly.

The only requirement after that is the casting of `Object` to `JavaExtension` or `ScalaExtension`.  I've pushed this into a new `SupportedLanguage` interface so that it's not possible to mismatch language and extension like `Conversions.toJavaExtension(gradleSourceSet.getExtensions().get(SupportedLanguages.SCALA));`

See what you think

Arthurm1:
@jdneo If you're happy with this refactoring then I'll refactor the Kotlin PR

model/src/main/java/com/microsoft/java/bs/gradle/model/JavaExtension.java
```diff
@@ -11,10 +11,6 @@
  * The extension model for Java language.
  */
 public interface JavaExtension extends Serializable {
```
jdneo: Do you think if it makes sense that creating an interface called something like `LanguageExtension` and has a method called `convert()`.

Each language extension extends this interface and implemented the `convert` by their own?

Arthurm1: @jdneo I've added another commit to push all conversions (and reflection use) into the LanguageExtension.

plugin/src/test/java/com/microsoft/java/bs/gradle/plugin/GradleBuildServerPluginTest.java
```diff
@@ -73,10 +75,9 @@ void testModelBuilder() throws IOException {
             dependency -> dependency.getModule().contains("gradle-api")
         ));
 
-        DefaultJavaExtension javaExtension = Conversions.toJavaExtension(
-            gradleSourceSet.getExtensions().get(SupportedLanguages.JAVA));
+        JavaExtension javaExtension =
```
jdneo: What about this: `JavaExtension javaExtension = (JavaExtension) gradleSourceSet.getExtensions()
.get(SupportedLanguages.JAVA.getBspName());`

I'm thinking that by writing like this, we can limit the convert interface only exist in `LanguageExtension`, which maybe makes it simpler?

Arthurm1: @jdneo Are you reviewing since the new commit - Github seems to think so but I can't tell for sure?
I've deleted the `Conversions` class and moved all this into the `LanguageExtension`.
I can squash the commits if you want

jdneo: I think I'm reviewing the latest one.

I left above comment because I see both `LanguageExtension` and `SupportedLanguage` has declares `convert()` method.

Arthurm1: Oh I see.  The reason I've put it in `SupportedLanguage` as well is that then you never need to cast it.
So no one can mismatch the types and names e.g.
`JavaExtension javaExtension = (JavaExtension) gradleSourceSet.getExtensions() .get(SupportedLanguages.SCALA.getBspName());`
or
`ScalaExtension javaExtension = (ScalaExtension) gradleSourceSet.getExtensions() .get(SupportedLanguages.JAVA.getBspName());`
Otherwise there's a lot of typecasting in the tests.  This way it's impossible to miscast.

jdneo: I see. What about giving it another name, maybe something like `getExtension()` or `getLanguageExtension()`? And give some document explaining the purpose of this method?

I'm trying to avoid introduce confusion between those two `convert()` method.

Arthurm1: @jdneo

I've renamed the `convert` method to `getExtension` and added a description.

I've also made it even simpler to get a language extension as you can just pass the sourceset...
```ScalaExtension scalaExtension = SupportedLanguages.SCALA.getExtension(sourceSet);```

I've also introduced `getLanguage` to the `LanguageModelBuilder` to prevent builders requesting the wrong `getLanguageCompileTask`

File: plugin/src/main/java/com/microsoft/java/bs/gradle/plugin/ScalaLanguageModelBuilder.java [hunk 5]
```diff
@@ -6,9 +6,21 @@
...
+import java.util.Optional;
+import java.util.Set;
 
+import com.microsoft.java.bs.gradle.model.GradleModuleDependency;
+import com.microsoft.java.bs.gradle.model.ScalaExtension;
+import com.microsoft.java.bs.gradle.model.SupportedLanguage;
+import com.microsoft.java.bs.gradle.model.impl.DefaultScalaExtension;
 import org.gradle.api.Project;
 import org.gradle.api.file.SourceDirectorySet;
+import org.gradle.api.internal.tasks.scala.MinimalScalaCompileOptions;
```
jdneo, line 21: Do you think it's necessary/possible to have some kind of fallback in case some breaking changes happens to those internal apis?

plugin/src/test/java/com/microsoft/java/bs/gradle/plugin/GradleBuildServerPluginTest.java
jdneo: New changes look good! Thank you!

plugin/src/main/java/com/microsoft/java/bs/gradle/plugin/ScalaLanguageModelBuilder.java
Arthurm1: I'd say that's more of a question across the whole codebase.  If you do a regex search on `\.gradle.*\.internal` you can see that internal apis are used in 5 different files.
From my experience with Gradle since around version 5, any changes to the Java/Scala APIs have been clearly deprecated for a few versions before being removed.
I guess part of it is how quickly you could respond to an API change by putting out a new version and having that picked up by VSCode.  There should be plenty of time with the deprecation warnings.
I'd like to see this project published to Maven so that other BSP clients (e.g. [Metals](https://github.com/scalameta/metals)) could easily install it and pick up any fixes.  AFAIK at the moment the only way to install this BSP server is to install the VSCode extension but lots of people use other UIs.  Could you look to publishing it on Maven?  I could raise an issue but it's not something I could submit a PR for.  There are lots of microsoft java projects up on Maven so maybe one of your colleagues can help.

jdneo: > I guess part of it is how quickly you could respond to an API change by putting out a new version and having that picked up by VSCode. There should be plenty of time with the deprecation warnings.

Makes sense!

> Could you look to publishing it on Maven?

Filed an issue: https://github.com/microsoft/build-server-for-gradle/issues/129, I'll discuss this with my team.

jdneo (approved):
LGTM. This PR also nicely addresses the issue(https://github.com/microsoft/build-server-for-gradle/issues/128), which is very great!

PR #120, opened 2023-12-14
Add Kotlin support

Arthurm1:
Adds Kotlin support.

I don't use Kotlin but I thought it would be good to get support for a few languages to help decide on best design for handling multiple languages.

BSP spec is not finalised for Kotlin.  See [here](https://github.com/build-server-protocol/build-server-protocol/issues/520).  I've no idea about how to setup `associates` in a Gradle build so I've ignored it.

Kotlin plugin doesn't seem to be part of Gradle so everything here is done with reflection - I guess you could add the Kotlin library into the `plugin` project if you don't want to use reflection.

Up to you if you want to merge. I could change this to a draft PR.

### DIFF
--- a/server/src/main/java/com/microsoft/java/bs/core/internal/managers/BuildTargetManager.java
+++ b/server/src/main/java/com/microsoft/java/bs/core/internal/managers/BuildTargetManager.java
@@ -18,11 +18,13 @@ import java.util.stream.Collectors;
 import ch.epfl.scala.bsp4j.JvmBuildTarget;
 import ch.epfl.scala.bsp4j.ScalaBuildTarget;
 import ch.epfl.scala.bsp4j.ScalaPlatform;
+import ch.epfl.scala.bsp4j.extended.KotlinBuildTarget;
 import com.microsoft.java.bs.core.internal.model.GradleBuildTarget;
 import com.microsoft.java.bs.gradle.model.BuildTargetDependency;
 import com.microsoft.java.bs.gradle.model.GradleSourceSet;
 import com.microsoft.java.bs.gradle.model.GradleSourceSets;
 import com.microsoft.java.bs.gradle.model.JavaExtension;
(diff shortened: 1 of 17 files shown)


server/src/main/java/com/microsoft/java/bs/core/internal/managers/BuildTargetManager.java
```diff
@@ -109,17 +122,49 @@ private List<String> getBuildTargetTags(boolean hasTests) {
     return tags;
   }
 
-  private void setJvmBuildTarget(GradleSourceSet sourceSet, BuildTarget bt) {
+  private void setBuildTarget(GradleSourceSet sourceSet, BuildTarget bt) {
+    if (sourceSet.isKotlin()) {
```
jdneo: If a source set applies kotlin, scala & java plugins, then which one should be returned as the data of the build target?

I've not idea about this. I asked this question in the BSP discord channel.

Arthurm1: I've no idea either.  I doubt it's possible.  I think that both the Scala and Kotlin plugins are Java aware and able to mix in their languages with Java so you could have Java classes referencing Scala classes and vice versa in the same project.  But neither plugin is aware of the other language so Scala wouldn't know what to do with Kotlin and Kotlin wouldn't know what to do with Scala.  Happy to be corrected on this but I think the code would have to be in separate projects.

jdneo: Searched in google and found that there are some people doing that: https://discuss.gradle.org/t/kotlin-groovy-and-java-compilation/14903/8.

But I guess that's a very rare case.

---

Update:

Looks like BSP team is targeting to address this in 3.0: https://github.com/build-server-protocol/build-server-protocol/discussions/487

Basically, allowing multiple build target data.

Arthurm1:
rebased

PR #123, opened 2023-12-21
Change display name format

Arthurm1:
Changes display name format to use `projectName [sourceset]`.

The previous way of creating a unique display name was getting too complicated when factoring in various build types e.g. nested projects, composite builds etc.


jdneo (approved):

jdneo:
Thank you @Arthurm1!

PR #124, opened 2023-12-23
Support running main classes, tests and test discovery

Arthurm1:
Implements BSP test discovery.  Only returns the classes but BSP doesn't seem to allow test discovery down to the method level.

Requires Gradle >= 8.3 as `--test-dry-run` has only recently been added.  There doesn't appear to be any other way to discover tests.

This is what this project looks like in VSCode test explorer
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,9 @@ Following BSP requests are supported in the current implementation:
 - `buildTarget/cleanCache`
 - `buildTarget/javacOptions`
 - `buildTarget/scalacOptions`
+- `buildTarget/jvmTestEnvironment`
+- `buildTarget/run`
+- `buildTarget/test`
 - `workspace/buildTargets`
 - `workspace/reload`
 
--- a/model/src/main/java/com/microsoft/java/bs/gradle/model/GradleSourceSet.java
+++ b/model/src/main/java/com/microsoft/java/bs/gradle/model/GradleSourceSet.java
@@ -111,6 +111,11 @@ public interface GradleSourceSet extends Serializable {
   public boolean hasTests();
 
   /**
+   * list of test tasks that are associated with this source set.
+   */
+  public Set<GradleTestTask> getTestTasks();
+
+  /**
    * Extensions of the source set.
    */
   public Map<String, LanguageExtension> getExtensions();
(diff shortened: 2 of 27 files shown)


Arthurm1:
Added support for `buildTarget/test` so tests can now be run at the class level.

I've also changed `BuildTargetServerIntegrationTest` to create a client for recording of server messages e.g. test reports.

Arthurm1:
Added support for buildTarget/run

Arthurm1:
rebased

jdneo:
Hi @Arthurm1,

some rebase work might need for this PR. Before that, may I ask the reason why you want the BS to support test discovery?

(I haven't look through the change, but looks like test discovery is not covered in BSP, is it correct?)

Arthurm1:
rebased

> (I haven't look through the change, but looks like test discovery is not covered in BSP, is it correct?)

Test discovery is a JVM only extension: https://build-server-protocol.github.io/docs/extensions/jvm#buildtargetjvmtestenvironment-request

Metals can do test discovery on its own on the currently open file but it's slightly painful as it has to know how to use each test framework and there are many: JUnit, TestNG, ScalaTest, MUnit, ScalaCheck, Specs2...
Even with that, it wouldn't be able to discover any [custom defined tests](https://docs.gradle.org/current/samples/sample_jvm_multi_project_with_additional_test_types.html) in Gradle

jdneo:
My thought about the test support is that we can do it progressively.

So the first step can be:

- client is responsible for test discovery.
- bs is used to resolve the classpath and other information that will be used by client to launch the test execution.

The next step can be the discovery support (considering currently the test discovery only support the class level). WDYT?


And if possible, separate the the run and test support in different PRs would definitely help a lot :)

Arthurm1:
So you want me to split into 3 PRs?  Test Discovery, Test Run, Main Run?

jdneo:
> So you want me to split into 3 PRs? Test Discovery, Test Run, Main Run?

Thank you. If you don't mind to do the split, That would be great :)

And I hope the order could be `Test Run, Main Run, Test Discovery`. Due to that the Test delegation is frequently asked from the community in VS Code Java.

PR #125, opened 2023-12-26
refactor: Support multiple languages.

jdneo:


jdneo:
@Arthurm1 A draft for multi-language support refactor. It's not finished yet, I've roughly done the plugin part. Would like to hear your thoughts before I am going on.

Basically, different language contributors can contribute.
- source dir
- generated source dir
- extenion(language specific data, like java version, source level, target level for java) to the `GradleSourceSet`

Arthurm1:
Looks fine to me.

I'd wonder about eventually moving JavaHome/Version/GradleVersion from...
```
    extension.setJavaHome(DefaultInstalledJdk.current().getJavaHome());
    extension.setJavaVersion(DefaultInstalledJdk.current().getJavaVersion().getMajorVersion());
    extension.setGradleVersion(project.getGradle().getGradleVersion());
```
into `SourceSetsModelBuilder`.

If BSP 3.0 has a map of `DataKind -> Data` for build targets then maybe you could introduce a `GradleDataKind` instead of extending the `JvmBuildTarget`.  I think it would be handy for clients to know something about the BSP Server even if it's only just info that they can report to users.  Definitely handy for the user to know what version of Gradle they're attached to as they may need a certain version for certain parts of BSP to work.  But maybe that's for later with BSP 3.0.

I assume `SourceSetModelBuilder` is to be deleted - it doesn't seem to be used anymore?

I don't think it matters too much if the refactor is not perfect - as it's all internal API it can be changed if necessary.

jdneo:
@Arthurm1 I think I'm close to finish the refactoring stuff. If you have time to review, it would be greatly appreciated!

plugin/src/test/java/com/microsoft/java/bs/gradle/plugin/Conversions.java
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * Utils to convert objects between two different class loaders to avoid
+ * ClassCastException.
+ */
+public class Conversions {
+
+  /**
+   * Convert an object to a DefaultJavaExtension.
+   */
+  public static DefaultJavaExtension toJavaExtension(Object object) {
```
Arthurm1: Why not do this copy in `DefaultGradleSourceSet` constructor like the other classes are copied?  Seems messy to use reflection.

model/src/main/java/com/microsoft/java/bs/gradle/model/JavaExtension.java
```diff
@@ -0,0 +1,48 @@
...
+
+  /**
+   * The java version this target is supposed to use.
+   */
+  public String getJavaVersion();
+
+  /**
+   * The Gradle version of the project.
+   */
+  public String getGradleVersion();
```
Arthurm1: Should this be in `GradleSourceSet`?  It's common to all languages.

model/src/main/java/com/microsoft/java/bs/gradle/model/impl/DefaultGradleSourceSet.java
```diff
@@ -39,30 +39,18 @@ public class DefaultGradleSourceSet implements GradleSourceSet {
...
-
-  private List<String> compilerArgs;
-
   private Set<GradleModuleDependency> moduleDependencies;
 
   private Set<BuildTargetDependency> buildTargetDependencies;
 
   private boolean hasTests;
 
+  private Map<String, Object> extensions;
```
Arthurm1, line 54: It might be clearer to have a `LanguageExtension` interface that all the extensions inherit from, even if it is blank.  Then you're not using `Object` in collections and methods.

plugin/src/main/java/com/microsoft/java/bs/gradle/plugin/SourceSetCache.java
```diff
@@ -0,0 +1,55 @@
...
+  public Collection<DefaultGradleSourceSet> getAllGradleSourceSets() {
+    return this.gradleSourceSetMapping.values();
+  }
+
+  public Collection<SourceSet> getAllSourceSets() {
+    return this.gradleSourceSetMapping.keySet();
+  }
+
+  public Collection<Project> getAllProjects() {
+    return this.projectMapping.values();
```
Arthurm1: Will this return duplicates when there are multiple sourcesets per project?

server/src/main/java/com/microsoft/java/bs/core/internal/managers/BuildTargetManager.java
```diff
@@ -110,13 +112,21 @@ private List<String> getBuildTargetTags(boolean hasTests) {
   }
 
   private void setJvmBuildTarget(GradleSourceSet sourceSet, BuildTarget bt) {
+    DefaultJavaExtension javaExtension = ConversionUtils.toJavaExtension(
+        sourceSet.getExtensions().get("java"));
```
Arthurm1: Use `JavaLanguageModelBuilder#LANGUAGE_ID` instead of "java"?

plugin/src/main/java/com/microsoft/java/bs/gradle/plugin/GradleBuildServerPlugin.java
```diff
@@ -13,15 +16,36 @@
...
   public void apply(Project project) {
     registry.register(new SourceSetsModelBuilder());
   }
+
+  private void registerSupportedLanguages() {
+    String supportedLanguagesProps = System.getProperty("bsp.gradle.supportedLanguages");
+    if (supportedLanguagesProps != null) {
+      String[] supportedLanguages = supportedLanguagesProps.split(",");
+      for (String language : supportedLanguages) {
+        if (language.equalsIgnoreCase("java")) {
```
Arthurm1: Use `JavaLanguageModelBuilder#LANGUAGE_ID` instead of "java"?

plugin/src/main/java/com/microsoft/java/bs/gradle/plugin/GradleBuildServerPlugin.java
```diff
@@ -13,15 +16,36 @@
...
+          SUPPORTED_LANGUAGE_BUILDERS.add(new JavaLanguageModelBuilder());
+        } else if (language.equalsIgnoreCase("scala")) {
```
Arthurm1: Use `ScalaLanguageModelBuilder#LANGUAGE_ID` instead of "scala"?

server/src/main/java/com/microsoft/java/bs/core/internal/services/BuildTargetService.java
```diff
@@ -295,7 +297,14 @@ public JavacOptionsResult getBuildTargetJavacOptions(JavacOptionsParams params)
       }
 
       GradleSourceSet sourceSet = target.getSourceSet();
-      List<String> classpath = sourceSet.getCompileClasspath().stream()
+      DefaultJavaExtension javaExtension = ConversionUtils.toJavaExtension(
+          sourceSet.getExtensions().get("java"));
```
Arthurm1: Use `JavaLanguageModelBuilder#LANGUAGE_ID` instead of "java"?

server/src/main/java/com/microsoft/java/bs/core/internal/utils/ConversionUtils.java
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * Utils to convert objects between two different class loaders to avoid
+ * ClassCastException.
+ */
+public class ConversionUtils {
+
+  /**
+   * Convert an object to a DefaultJavaExtension.
+   */
+  public static DefaultJavaExtension toJavaExtension(Object object) {
```
Arthurm1: Same as in `Conversions` - why not do this copy in DefaultGradleSourceSet constructor like the other classes are copied?

Arthurm1:
@jdneo Looks good - I've added some minor comments

plugin/src/test/java/com/microsoft/java/bs/gradle/plugin/Conversions.java
jdneo: If without this util. To get a java extension, for example, I need to write:

```java
DefaultJavaExtension javaExtension = (DefaultJavaExtension) gradleSourceSet.getExtensions().get("java");
```

Then when I run the unit tests, I got `ClassCastException` saying that the classes are loaded by different class loaders.

I haven't dig deep into this issue, the solution(workaround) I found is to use such util to do a conversion.

model/src/main/java/com/microsoft/java/bs/gradle/model/impl/DefaultGradleSourceSet.java
jdneo: Makes sense!

server/src/main/java/com/microsoft/java/bs/core/internal/managers/BuildTargetManager.java
jdneo: The project `plugin` is not on the classpath of the `server` project.

Put this const somewhere in the `model` project?

model/src/main/java/com/microsoft/java/bs/gradle/model/impl/DefaultGradleSourceSet.java
jdneo: I'm stuck here.

After change to `LanguageExtension` interface, the return of `gradleSourceSet.getExtensions().get("java")` becomes a proxy of `LanguageExtension`.

I cannot find a way to convert it to `JavaExtension`. class cast and reflection are not working.

----

One way is to use `ProtocolToModelAdapter#unpack()` to get the wrapped object and then do the reflection conversion. But `ProtocolToModelAdapter` is internal api, which is not ideal.

Arthurm1: Ah - no problem then, I'd leave as is

server/src/main/java/com/microsoft/java/bs/core/internal/managers/BuildTargetManager.java
Arthurm1: Oops - I thought it was in Model, not Plugin.  Up to you - it's not a big issue

model/src/main/java/com/microsoft/java/bs/gradle/model/impl/DefaultGradleSourceSet.java
jdneo: But I think your suggestion makes sense. I'll file an issue for this debt after this PR is merged.

Arthurm1:
@jdneo Are there more changes needed for this or are you planning on merging soon?

jdneo:
image

@testforstephen, sorry to introduce such large change. This refactoring is to make the multi-language support easier. Each language should implement a `LanguageModelBuilder` which is used to provide specific language information to the SourceSetModelBuilder.

One thing to mention that, the return value of `getExtensionFor()` depends on the [BSP defined extension](https://build-server-protocol.github.io/docs/extensions/java) for different languages.

jdneo:
> Are there more changes needed for this or are you planning on merging soon?

@Arthurm1 I'll try to make sure it be merged recently.

server/src/main/java/com/microsoft/java/bs/core/internal/managers/BuildTargetManager.java
testforstephen: It's better to define the supported language id as constants for reuse in model, plugin and server.

server/src/main/java/com/microsoft/java/bs/core/internal/utils/ConversionUtils.java
```diff
@@ -0,0 +1,45 @@
...
+import java.lang.reflect.InvocationTargetException;
+import java.util.List;
+
+import com.microsoft.java.bs.gradle.model.impl.DefaultJavaExtension;
+
+/**
+ * Utils to convert objects between two different class loaders to avoid
+ * ClassCastException.
+ */
+public class ConversionUtils {
```
testforstephen: this is related to the model logic, why not to move this utility to the model layer?

testforstephen (approved):

model/src/main/java/com/microsoft/java/bs/gradle/model/impl/DefaultGradleSourceSet.java
jdneo: Filed: https://github.com/microsoft/build-server-for-gradle/issues/128

PR #131, opened 2024-04-01
Initial Groovy support

Arthurm1:
BSP doesn't have any specifics about Groovy so the details in the language model are just really the source directories.

I guess users can request more info if they need e.g. groovyCompileOptions but that would need BSP changes.

For now this will allow standard BSP commands to be used on groovy projects e.g. compile, test, run etc.


graemerocher:
@jdneo this would be a great addition as well once the delegating to Gradle for test/run is done. Currently if you have a Java project with Spock tests you can't run them

jdneo:
I'm not sure if this can help with the Spock test. Main problem is that the JDT language server does not support groovy right now, so the test discovery might be a problem.

Delegate the test discovery to Build Server as well might be a solution.

Anyway, some investigation is needed here.

PR #132, opened 2024-04-04
Add jdk 22 compatibility

Arthurm1:
Gradle supports JDK 22 from version 8.7

https://docs.gradle.org/current/userguide/compatibility.html


jdneo:
Since the column `Support for running Gradle` is still N/A, I think we need to wait more...

Arthurm1:
Ah - my mistake

jdneo:
https://docs.gradle.org/current/userguide/compatibility.html

8.8 supports java 22. Would you like to reopen this ?

jdneo:
Never mind, I'll take it: https://github.com/microsoft/build-server-for-gradle/pull/152

PR #133, opened 2024-04-04
Ignore compilation on empty build targets

Arthurm1:
There's no `alive` check in BSP so the client can repeatedly check that the server is still up and running.

Metals uses `buildTarget/compile` for this with an empty list of targets but this causes a complete refresh of all the source sets.

So this adds a guard so no refresh is done if the list of targets is empty.


jdneo (approved):

jdneo:

PR #134, opened 2024-04-04
Update project to use Gradle 8.7

Arthurm1:
Updates Gradle from 8.1 to 8.7 so jdk 21 can be used to develop

I've left the buildship version as is so some tests will fail unless JDK 17 or lower is used to run them.  Can this also be changed or is it set to 7.4.2 for a reason...
https://github.com/microsoft/build-server-for-gradle/blob/ec53a53b8b2b66cad5303e09474010b39d9303ff/server/src/main/java/com/microsoft/java/bs/core/internal/gradle/Utils.java#L45


jdneo:
They are some stale(invalid) checks. I'll raise a PR to remove them.

jdneo:
Removed in https://github.com/microsoft/build-server-for-gradle/pull/137

jdneo:
> I've left the buildship version as is so some tests will fail unless JDK 17 or lower is used to run them.

Which tests?

Arthurm1:
> Which tests?

`testJavaCompilerArgsToolchain` failed for me but I think that's because I was switching between jdk-17 and jdk-21.  All work now.

jdneo (approved):

jdneo:

PR #135, opened 2024-04-04
Shift sourceset population out of buildInitialize

Arthurm1:
Metals (BSP client) expects a prompt response to `build/initialize` and if it doesn't get one within 30 seconds it assumes the build server has not started correctly and so attempts to restart it.

Large projects take longer than this because the sourceset population takes a long time.

This shifts the population of sourcesets away from `build/initialize` and into `workspace/buildTargets`

It also shifts the responsibility of handling `workspace/reload` away from `LifecycleService` and into `BuildTargetService`.  The `LifecycleService` code was similar to that `RefetchBuildTargetTask` so this allows doing away with `RefetchBuildTargetTask` and also means that any call to `workspace/reload` will cause a `buildTarget/didChange` event if the workspace has changed - which wasn't the case before.

### DIFF
--- a/server/src/main/java/com/microsoft/java/bs/core/Launcher.java
+++ b/server/src/main/java/com/microsoft/java/bs/core/Launcher.java
@@ -48,8 +48,7 @@ public class Launcher {
     BuildTargetManager buildTargetManager = new BuildTargetManager();
     PreferenceManager preferenceManager = new PreferenceManager();
     GradleApiConnector connector = new GradleApiConnector(preferenceManager);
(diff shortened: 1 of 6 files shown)


jdneo (approved):
LGTM! I think this is the correct way to move forward. initialization should be as fast as possible.

@Arthurm1 could you please resolve the conflicts? Then I'll merge it.

Arthurm1:
@jdneo rebased

jdneo:
Thank you @Arthurm1

PR #136, opened 2024-04-04
add message checking to integration tests

Arthurm1:
Currently the integration tests only handle direct replies to BSP messages e.g. `workspace/buildTargets` returns `WorkspaceBuildTargetsResult`

This PR allows monitoring of async messages e.g. `build/taskProgress` and adds tests for these.

It's the first step in testing progress reports per build target and eventually diagnostic reports.

It also gets rid of the global `Launcher.client` which I think is a good thing.

Also fixed: the `onBuildTaskProgress` notification has a `compile-task` data kind so `data` must contain a `CompileTask`

### DIFF
--- a/server/src/test/java/com/microsoft/java/bs/core/internal/server/BuildTargetServerIntegrationTest.java
+++ b/server/src/test/java/com/microsoft/java/bs/core/internal/server/BuildTargetServerIntegrationTest.java
@@ -4,11 +4,22 @@
 package com.microsoft.java.bs.core.internal.server;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.io.File;
+import java.io.IOException;
+import java.io.PipedInputStream;
+import java.io.PipedOutputStream;
 import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.function.BiConsumer;
 import java.util.stream.Collectors;
 
 import ch.epfl.scala.bsp4j.DependencyModulesParams;
@@ -19,7 +30,6 @@ import ch.epfl.scala.bsp4j.MavenDependencyModule;
 import ch.epfl.scala.bsp4j.MavenDependencyModuleArtifact;
 import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.BeforeAll;
-import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
 import com.microsoft.java.bs.core.Launcher;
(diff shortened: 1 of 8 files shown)


jdneo:
Had a quick look and the idea looks good to me.

Could you please resolve the conflict?

Arthurm1:
@jdneo Done

Arthurm1:
rebased

server/src/test/java/com/microsoft/java/bs/core/internal/server/BuildTargetServerIntegrationTest.java
```diff
@@ -53,55 +184,122 @@ static void beforeClass() {
...
-  @AfterAll
-  static void afterClass() {
-    System.clearProperty(Launcher.PROP_PLUGIN_DIR);
+  @AfterEach
+  void afterEach() {
+    // closing the streams and pool shuts down the servers
+    try {
+      clientOut.close();
+      /*
+         Gradle will still fire messages for a while after the connection is closed
```
jdneo: Should we call the `.buildShutdown()` for the server? Will this help with the output stream issue mentioned here?

Arthurm1: I'd forgotten that existed.  I've added that in and refactored so that the streams are closed using try-with-resources.  I now can't reproduce the issue I had before.

jdneo (approved):
LGTM

jdneo:
Thank you @Arthurm1!

server/src/test/java/com/microsoft/java/bs/core/internal/server/BuildTargetServerIntegrationTest.java
```diff
@@ -53,79 +176,157 @@ static void beforeClass() {
...
+      assertTrue(cleanResult.getCleaned());
+      client.waitOnMessages(2, 2, 1);
+      client.clearMessages();
+
+      // compile targets
+      CompileParams compileParams = new CompileParams(btIds);
+      CompileResult compileResult = gradleBuildServer.buildTargetCompile(compileParams).join();
+      assertEquals(StatusCode.OK, compileResult.getStatusCode());
+      client.waitOnMessages(2, 2, 1);
+      BuildTargetIdentifier testBt = getBt(btIds, "/junit5-jupiter-starter-gradle/", "test");
```
jdneo, line 325: @Arthurm1 Does this case pass on your side? Due to the current implementation at: https://github.com/microsoft/build-server-for-gradle/blob/59b80ec5d8d29174ada027449ab330483e656b8f/server/src/main/java/com/microsoft/java/bs/core/internal/gradle/GradleApiConnector.java#L121-L126, it should fail.

It fails on my machine, but I have no idea why it passes in GitHub Actions.

Arthurm1: @jdneo Yes it passes on mine.  What line does it fail on - 324?  The messages I have at that point are...
image
It's possible you have additional messages.  I noticed that the `CompletableFuture.runAsync(this::reloadWorkspace);` triggered after a compile will cause more messages to be sent to the client but that's normally too slow to register.  I have fix for this and can submit a PR, but it's only ever happened to me when I've sat on a breakpoint for a while.

jdneo: I suspect this limitation causes the failure:

Target id for both `main` and `test` are sent to be compiled, and the task `classes` and `testClasses` are executed in one invocation. But because of https://github.com/microsoft/build-server-for-gradle/blob/59b80ec5d8d29174ada027449ab330483e656b8f/server/src/main/java/com/microsoft/java/bs/core/internal/gradle/GradleApiConnector.java#L126

only `main` target id is set to the compile reporter, which fails line 326?

Arthurm1: @jdneo This makes sense - it could be failing on your machine due to message ordering differences.  I've finished off #116 which should fix this if you'd like to review it.  It reports on progress-per-target rather than just reporting everything on the first target.

jdneo: Cool! We can eliminate a protocol violation then. I'll review it these days. Thank you!

PR #139, opened 2024-04-14
Add more error reporting for getGradleSourceSets failure

Arthurm1:
This passes `stderr` back if retrieving source sets fails.  Before there sometimes wasn't enough information to diagnose.

### DIFF
--- a/server/src/main/java/com/microsoft/java/bs/core/internal/gradle/GradleApiConnector.java
+++ b/server/src/main/java/com/microsoft/java/bs/core/internal/gradle/GradleApiConnector.java
@@ -72,10 +72,12 @@ public class GradleApiConnector {
       throw new IllegalStateException("Failed to get init script file.");
     }
     TaskProgressReporter reporter = new TaskProgressReporter(new DefaultProgressReporter());
+    ByteArrayOutputStream errorOut = new ByteArrayOutputStream();
     String summary = "";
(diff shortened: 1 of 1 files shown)


server/src/main/java/com/microsoft/java/bs/core/internal/gradle/GradleApiConnector.java
```diff
@@ -93,10 +96,10 @@ public GradleSourceSets getGradleSourceSets(URI projectUri) {
       // since the model returned from Gradle TAPI is a wrapped object, here we re-construct it
       // via a copy constructor and return as a POJO.
       return new DefaultGradleSourceSets(customModelBuilder.get());
-    } catch (GradleConnectionException | IllegalStateException e) {
-      summary = e.getMessage();
+    } catch (GradleConnectionException | IllegalStateException | IOException e) {
+      summary = e.getMessage() + "\n" + errorOut.toString();
```
jdneo: Let's check if errorOut is empty or not before appending to the summary text.

Arthurm1: Done

jdneo (approved):
LGTM

jdneo:
Thank you @Arthurm1

PR #143, opened 2024-05-14
Auto download toolchain in test project

Arthurm1:
Downloads jdk 17 toolchain if it's not setup locally.

Only for toolchain test

### DIFF
--- a/testProjects/java-compilerargs-toolchain/settings.gradle
+++ b/testProjects/java-compilerargs-toolchain/settings.gradle
@@ -1,1 +1,4 @@
+plugins {
+    id("org.gradle.toolchains.foojay-resolver-convention") version "0.8.0"
+}
 rootProject.name = 'java-compilerargs-toolchain'


jdneo:
I haven't dig deep into that issue. But I'm curious about the root cause. Do you have more information?

Arthurm1:
AFAIK Gradle will not download JDKs automatically if it can't find them.  It recommends some 3rd party plugins for this: https://docs.gradle.org/current/userguide/toolchains.html#sec:provisioning

It's possible to install JDKs manually and have Gradle recognise them but I've had issues with this on Windows as I think it requires windows registry settings to point it to the JDK locations.  I could be wrong on this.

The CI azure pipeline specifies jdk 17 and since Gradle is running on that it can find JDK 17.  I imagine if the CI build moved to JDK 21 then this test would fail.  This PR should fix that issue although the CI checks may take slightly longer to run.

jdneo:
I see.

Instead of adding this 3rd party plugin. What about limit the related test cases to only run on java 17 using the annotation `@EnabledOnJre({JRE.JAVA_17})`?

Arthurm1:
I'm easy - whichever you prefer

jdneo:
Hmm, thought twice. Consider the java version updates over time, the JDK used in CI pipelines will be updated as well. The benefit of using the plugin is that in the future there is no need to update the test cases any more.

I'll go with your proposal.

jdneo (approved):
LGTM

jdneo:
Thank you @Arthurm1!

PR #144, opened 2024-05-24
Add support for running tests

Arthurm1:
Allows running of tests using `buildTarget/test` BSP request

There appears to be no JVM specific `TestParamsDataKind` defined in the BSP spec, only Scala based ones.  These should be fine to use on any JVM language (Java, Scala, Kotlin etc.)

The following `TestParamsDataKind` types are supported...
[`scala-test`](https://build-server-protocol.github.io/docs/extensions/scala#scalatestparams) which allows the client to run a list of tests at the class level per build target.
[`scala-test-suites-selection`](https://build-server-protocol.github.io/docs/extensions/scala#scalatestsuites) which allows the client to run a list of tests at the method level in a list of classes but only for 1 build target.

There are a number of caveats to testing...
1. Gradle >= 2.6 is required to run tests
2. Gradle >= 3.5 is required if environment variables are passed
3. `TaskStartParams` and `TaskFinishParams` do not contain a `Location` as Gradle only supplies the `suite`, `class`, `method` for each test, not the file uri.  That information is then lost as BSP has no way of reporting location except by file uri.
4. BSP has no structured way of reporting back the stacktrace on failed tests so that is done in the `TestFinish#message`


plugin/src/main/java/com/microsoft/java/bs/gradle/plugin/SourceSetsModelBuilder.java
```diff
@@ -76,9 +76,6 @@ public Object buildAll(String modelName, Project rootProject) {
         gradleSourceSet.setCleanTaskName(cleanTaskName);
         Set<String> taskNames = new HashSet<>();
         gradleSourceSet.setTaskNames(taskNames);
-        taskNames.add(classesTaskName);
-        taskNames.add(cleanTaskName);
-        taskNames.add(getFullTaskName(projectPath, sourceSet.getProcessResourcesTaskName()));
```
jdneo, line 81: Is this on purpose to only monitor the compile task?

server/src/main/java/com/microsoft/java/bs/core/internal/reporter/TestReportReporter.java
```diff
@@ -0,0 +1,147 @@
...
+          if (event instanceof StartEvent) {
+            TaskStartParams startParam = new TaskStartParams(taskId);
+            startParam.setMessage("Start test");
+            startParam.setDataKind("test-start");
+            startParam.setEventTime(event.getEventTime());
+            TestStart testStart = new TestStart(event.getDisplayName());
+            // TODO Gradle does not provide file/position data
+            startParam.setData(testStart);
+            client.onBuildTaskStart(startParam);
+          } else if (event instanceof FinishEvent) {
```
jdneo: nit: the enhanced instanceof pattern introduced in Java 14 may simplify the code just a little bit.

server/src/main/java/com/microsoft/java/bs/core/internal/reporter/TestReportReporter.java
```diff
@@ -0,0 +1,147 @@
...
+        // only report on methods
+        JvmTestOperationDescriptor descriptor = (JvmTestOperationDescriptor) event.getDescriptor();
+        if (descriptor.getClassName() != null && descriptor.getMethodName() != null) {
+          TaskId taskId = getTaskId(descriptor.getDisplayName());
+          if (event instanceof StartEvent) {
+            TaskStartParams startParam = new TaskStartParams(taskId);
+            startParam.setMessage("Start test");
+            startParam.setDataKind("test-start");
+            startParam.setEventTime(event.getEventTime());
+            TestStart testStart = new TestStart(event.getDisplayName());
```
jdneo: Looks like the `event.getDisplayName()` will report event like:

```
Test isBasicTest()(com.example.project.PassingTests) started
```

which contains the class name and method name that the client can use it to target the test item. But when it comes to the case that the test case is annotated by `@DisplayName()`, or parameterized test whose input is a object:

```java
  class User {}

  static Stream<Arguments> users() {
    return Stream.of(Arguments.of(new PassingTests().new User()));
  }

  @ParameterizedTest
  @MethodSource("users")
  void testCheckUser(User user) {
    assertTrue(true);
  }
```

the method name information is lost.

What do you think about we introduced a structured message, the message contains not only the display name of the event, but also the class name and method name from the descriptor. Then the client can use them when needed.

plugin/src/main/java/com/microsoft/java/bs/gradle/plugin/SourceSetsModelBuilder.java
Arthurm1: I keep going back and forth on this.  It shouldn't matter if more are included but I think they're meaningless in general reporting of compile progress and warning/error count.
It's easy to put them back in but I think it's cleaner without them.
You can see further down that a compile task per individual language will be added.

server/src/main/java/com/microsoft/java/bs/core/internal/reporter/TestReportReporter.java
Arthurm1: I agree - do the VSCode java tools all require jdk 17 now?  I wasn't sure if 11 is still supported (to run under)

server/src/main/java/com/microsoft/java/bs/core/internal/reporter/TestReportReporter.java
Arthurm1: I think BSP isn't perfectly defined here.  It has `Location` which is a position in a file.  The test frameworks that I've seen (`junit`, `sbt/test-interface`) don't provide this info. Instead they provide `className` & `methodName`.
1. We could extends the `TestStart` & `TestFinish` classes to include `className` & `methodName`.
2. We could have a custom `TestFinishDataKind` to hold `className` & `methodName` as there don't appear to be any `TestFinishDataKind` types but that apply only to `TestFinish`, not `TestStart`.
3. We could pass this data encoded as JSON in the `displayName`.
4. We could apply to change the BSP spec.

All seem like fudges except the 4th, but the client does need this info somehow and I'm happy to go down the custom message route.

plugin/src/main/java/com/microsoft/java/bs/gradle/plugin/SourceSetsModelBuilder.java
jdneo: I vote for without them as well :)

server/src/main/java/com/microsoft/java/bs/core/internal/reporter/TestReportReporter.java
jdneo: Yes, in VS Code Java, the build server will be launched using jre 17 now.

server/src/main/java/com/microsoft/java/bs/core/internal/reporter/TestReportReporter.java
jdneo: Thank you for the points listed here!

I thought about this question again, extending `TestStart` & `TestFinish` sounds better to me. It's a feasible solution if we want to change BSP spec. On the other hand, an JSON encoded message somehow breaks the semantic meaning of `displayname`.

Arthurm1:
@jdneo Added `TestStartEx` and `TestFinishEx`

server/src/main/java/com/microsoft/java/bs/core/internal/gradle/GradleApiConnector.java
```diff
@@ -138,6 +150,83 @@ public StatusCode runTasks(URI projectUri, ProgressReporter reporter, String...
...
+            }
+            for (Map.Entry<String, Set<String>> classesMethods : entry.getValue().entrySet()) {
+              if (classesMethods.getValue() != null && !classesMethods.getValue().isEmpty()) {
+                launcher.withJvmTestMethods(classesMethods.getKey(), classesMethods.getValue());
+              } else {
+                launcher.withJvmTestClasses(classesMethods.getKey());
+              }
+            }
+            if (args != null && !args.isEmpty()) {
+              launcher.withArguments(args);
```
jdneo: Not a big issue. The document says: `If not configured, null, or an empty list is passed, then the reasonable default will be used.` So maybe we can skip the null & empty check to make the code a little bit simpler?

Same to `setJvmArguments()` & `setEnvironmentVariables()`

jdneo (commented):
I will next try to integrate this into the java test runner extension and see if there is any gaps:

test runner delegates test request to GBS -> GBS reports the test result back -> test runner shows the test result.

This PR will be suspended temporarily during this progress. Thank you @Arthurm1 for your contribution!

server/src/main/java/com/microsoft/java/bs/core/internal/gradle/GradleApiConnector.java
Arthurm1: Good point.  I've changed it.

Arthurm1:
@jdneo I've raised this https://github.com/build-server-protocol/build-server-protocol/issues/674

I've gone for new data types as I don't think `suite` and `class` would be applicable to all languages but feel free to comment otherwise

server/src/main/java/com/microsoft/java/bs/core/internal/services/BuildTargetService.java
```diff
@@ -469,6 +477,86 @@ public ScalacOptionsResult getBuildTargetScalacOptions(ScalacOptionsParams param
...
+          for (String classNames : testClassesItem.getClasses()) {
+            classesMethods.put(classNames, null);
+          }
+          testClasses.put(testClassesItem.getTarget(), classesMethods);
+        }
+        statusCode = connector.runTests(entry.getKey(), testClasses, testParams.getJvmOptions(),
+            params.getArguments(), null, client, params.getOriginId(),
+            compileProgressReporter);
+      } else if ("scala-test-suites-selection".equals(params.getDataKind())) {
+        // ScalaTestSuites is for a list of classes + methods
```
jdneo, line 511: Looks like `ScalaTestParams` doesn't supply the specific BuildTarget as well. Why do we only check it here?

Arthurm1: You mean `ScalaTestParams` _does_ supply the build target (inside each `ScalaTestClassesItem`) and `ScalaTestSuites` doesn't?
I'm not sure why `ScalaTestParams` does supply the build target and `ScalaTestSuites` doesn't - seems like an oversight.
Do you mean we should be filtering the `ScalaTestClassesItem` against `entry.getValue`?  You could be right about that.  Want me to add it?
Also - I've been force pushing every update as a single commit - do you prefer additional commits - does that make it easier to review changes?

jdneo: My bad! I didn't realize that `ScalaTestClassesItem` contains the build target. Sorry for the confusion.

> Do you mean we should be filtering the ScalaTestClassesItem against entry.getValue?

No, I'm not sure about that. I'm not clear about the spec here. Since both TestParam and ScalaTestClassesItem specifies the build target, which one wins when they are different?

> do you prefer additional commits 

Sure! Additional commits makes it easier to review. I can squash all the commits when merge.

server/src/main/java/com/microsoft/java/bs/core/internal/reporter/TestReportReporter.java
```diff
@@ -0,0 +1,146 @@
...
+    exception = null;
+    testDuration = 0;
+  }
+
+  @Override
+  public void statusChanged(ProgressEvent event) {
+    if (client != null) {
+      if (event.getDescriptor() instanceof JvmTestOperationDescriptor descriptor) {
+        // only report on methods
+        if (descriptor.getClassName() != null && descriptor.getMethodName() != null) {
```
jdneo: I found that only report progress when class name and method name are not null might have some issue.

Considering:

```java
@TestFactory
Stream<DynamicContainer> testContainers() {
  return Stream.of(dynamicContainer("Container", List.of(
    dynamicTest("Test", () -> {assertTrue(false);})
  )));
}
```
The result hierarchy from IDEA is:

image

Corresponds to the output of gradle tooling api：

 ```
PassingTests                     // 'PassingTests started' [className: "PassingTests"] [methodName: null]
    |-- testContainers            // 'Test suite 'testContainers() started' [className: null] [methodName: null]
             |-- Container        // 'Container started' [className: null] [methodName: null]
                      |-- Test    // 'Test started' [className: "PassingTests"] [methodName: "testContainers()[1][1]"]
```

I'm still thinking that how to display this hierarchy via BSP. It looks to me that we need to take the `name` and `parent` field of the `descriptor` into consideration as well.

Comment (jdneo) [reply server/src/main/java/com/microsoft/java/bs/core/internal/reporter/TestReportReporter.java]: A better example:

```
@TestFactory
Collection<DynamicContainer> testContainer() {
    List<DynamicContainer> containers = new ArrayList<>();
    containers.add(
            dynamicContainer("First Container", List.of(
                    dynamicTest("First test of first container", () -> {
                    }),
                    dynamicTest("Second test of first container", () -> {
                    })
            ))
    );
    containers.add(
            dynamicContainer("Second Container", List.of(
                    dynamicTest("First test of second container", () -> {
                    }),
                    dynamicTest("Second test of second container", () -> {
                    })
            ))
    );
    return containers;
}
```

Which should be able to rendered like this in VS Code, or other clients:

image

Comment (jdneo) [reply server/src/main/java/com/microsoft/java/bs/core/internal/reporter/TestReportReporter.java]: I've got a feeling that the introduced new fields: `className` and `methodName` may not be suitable to present the test progress/result for the above example.

To summarize the current properties that we can use:

1. The client needs to know whether this test event is a start or finish. this can be achieved via the data kind: `test-start` / `test-finish`
2. For test finish event, the client needs to know the result of it. This can be achieved by the `TestStatus` in the data.
3. For failed cases, stacktrace can be stored into `message`. (No need to introduce new field `stacktrace` I think)
4. Duration of each test case is needed, this can be calculated by checking the `eventTime`
5. Then, for each test case, we need to use `displayName` to store the content displayed in the ui. Take the above case as example: `First Container` or `First test of first container`.
6. We need another field acts as the `test id` to let the client target the current test case, at the mean time, the field should be able to contain the hierarchy information for the test execution. Still taking the above example: use the `JvmTestOperationDescriptor.getName()` to get the current test name, and use `/` to present different level of the hierarchy. The hierarchy can be restored via `JvmTestOperationDescriptor.getParent()`. So, during the execution, we will get following ids:

`com.example.project.PassingTests`
`com.example.project.PassingTests/testContainer()`
`com.example.project.PassingTests/testContainer()/testContainer()[1]`
`com.example.project.PassingTests/testContainer()/testContainer()[1]/testContainer()[1][1]`
...

The question is where should we store the ids, using `taskId` or introduce another field? I prefer to use `taskId`.

WDYT?

Comment (Arthurm1) [reply server/src/main/java/com/microsoft/java/bs/core/internal/reporter/TestReportReporter.java]: I've added the test hierarchy into `taskId`.  I think it's fine for testing - to make sure you've got everything for reporting but I imagine someone could create a dynamic test with a `/` in the name which would cause issues with your parsing.
Might be better to have a `TestNameEx` type like...
```
class TestNameEx {
  String name;
  TestNameEx parent;
}
```
I'm still only reporting on test tasks where `methodName != null`.  Is that OK? Or do you need `TaskStart` and `TaskFinish` for suites and classes as well?

Comment (jdneo) [reply server/src/main/java/com/microsoft/java/bs/core/internal/reporter/TestReportReporter.java]: Since `TaskId` has a member `parent`. How about using that to present the hierarchy? Then we can avoid `a dynamic test with a / in the name` problem.

Comment (jdneo) [reply server/src/main/java/com/microsoft/java/bs/core/internal/reporter/TestReportReporter.java]: > I'm still only reporting on test tasks where methodName != null. Is that OK? Or do you need TaskStart and TaskFinish for suites and classes as well?

I prefer having start and finish for suite and classes as well, which is align with the test execution in gradle.

Comment (Arthurm1) [reply server/src/main/java/com/microsoft/java/bs/core/internal/reporter/TestReportReporter.java]: > Since `TaskId` has a member `parent`. How about using that to present the hierarchy? Then we can avoid `a dynamic test with a / in the name` problem.

`TaskId#parent` is a `List<String>` not another `TaskId` so you can't chain them.  I guess we could add encode the tests and put them in the `parent` as a list like...
```
TestName1:com.example.project.PassingTests
TestName2:testContainer()
TestName3:testContainer()[1]
TestName4:testContainer()[1][1]
550e8400-e29b-41d4-a716-446655440000   // <-- real parent task id
```

It would be a bit messy though.

I could put a hierarchy into `TestStartEx` and `TestFinishEx`?

Comment (jdneo) [reply server/src/main/java/com/microsoft/java/bs/core/internal/reporter/TestReportReporter.java]: How will the hierarchy look like in `TestStartEx`/`TestFinishEx`?

Comment (Arthurm1) [reply server/src/main/java/com/microsoft/java/bs/core/internal/reporter/TestReportReporter.java]: I've added `TestName` in a new commit.  See what you think, we can always roll it back.
`TaskId` no longer has the encoded test hierarchy.
`TestStartEx` and `TestFinishEx` no longer have `SuiteName`, `ClassName`, `MethodName`.
`TestName` is hierarchical and defined as...
```java
  String displayName;
  String suiteName;
  String className;
  String methodName;
  TestName parent;
```
So now you can pick and choose what to display in VSCode at each level in the hierarchy.  `displayName` seems good for most levels.

You'll also get additional `TestStartEx` and `TestFinishEx` messages as class level and sometimes suite level, depending on whether Gradle reports it for that class.

The test pass, skip, fail counts haven't changed as they still only count the methods run.

server/src/main/java/com/microsoft/java/bs/core/internal/reporter/TestReportReporter.java
```diff
@@ -132,19 +126,25 @@ public void statusChanged(ProgressEvent event) {
...
+                skippedCount += 1;
+              }
             } else if (result instanceof TestSuccessResult) {
-              successCount += 1;
+              if (descriptor.getMethodName() != null) {
+                successCount += 1;
+              }
             }
             TaskFinishParams finishParam = new TaskFinishParams(taskId, statusCode);
             finishParam.setMessage("Finish test");
```
jdneo, line 150: What about put the stacktrace to the message? Then we extend less properties

jdneo (commented):
Thank you @Arthurm1, I'll continue the integration work between this work and the java test runner extension.

BTW, I guess we need to update this proposal: https://github.com/build-server-protocol/build-server-protocol/issues/674 as well?

server/src/main/java/com/microsoft/java/bs/core/internal/reporter/TestReportReporter.java
```diff
@@ -92,28 +81,33 @@ private List<JvmTestOperationDescriptor> getTestStack(
     }
     // earlier check means that classname will always be found so can't have i < 0
     // reverse list order
-    List<JvmTestOperationDescriptor> result = new ArrayList<>(i + 1);
+    TestName testName = null;
     while (i >= 0) {
-      result.add(fullStack.get(i));
+      JvmTestOperationDescriptor desc = fullStack.get(i);
+      TestName currentTestName = new TestName(desc.getDisplayName(), desc.getSuiteName(),
```
jdneo: I got a feeling that we can reduce the extended property in `TestName` even more.

Basically, looks like we can merge suiteName, className and methodName to one name.

For example, we can first call `JvmTestOperationDescriptor.getJvmTestKind()`. If it's `SUITE`, then we use `JvmTestOperationDescriptor.getSuiteName()` as the name. Ortherwise, if it's `ATOMIC`, use `JvmTestOperationDescriptor.getMethodName()`, or `getClassName()` -> when `getMethodName()` is null, as the name.

It's just an idea but I'm not sure if it's still valid across different test frameworks. Just left this idea here first. I will check it more widely after I've finish the integration.

Arthurm1:
> Thank you @Arthurm1, I'll continue the integration work between this work and the java test runner extension.
> 
> BTW, I guess we need to update this proposal: [build-server-protocol/build-server-protocol#674](https://github.com/build-server-protocol/build-server-protocol/issues/674) as well?

I'll update it when we're completely happy with the structure here.

server/src/main/java/com/microsoft/java/bs/core/internal/reporter/TestReportReporter.java
Arthurm1: If I'm going to submit a BSP protocol change then I'd prefer to keep it separate.

server/src/main/java/com/microsoft/java/bs/core/internal/reporter/TestReportReporter.java
Arthurm1: I'll try and add `TestNG` and `ScalaTest` projects to the integ tests and see what they come up with.  Do you know of any other test engines around?

server/src/main/java/com/microsoft/java/bs/core/internal/reporter/TestReportReporter.java
jdneo: ok

server/src/main/java/com/microsoft/java/bs/core/internal/reporter/TestReportReporter.java
jdneo: We can first focus on junit 4, 5 and testng. Actually from the telemetry, junit 5 is kind of dominant.

There are lots of request for spock, but that's a groovy based one. Not our focus right now.

jdneo: Some test samples I used: https://github.com/microsoft/vscode-java-test/tree/main/test/test-projects/junit/src/test/java

server/src/test/java/com/microsoft/java/bs/core/internal/server/BuildTargetServerIntegrationTest.java
```diff
@@ -370,6 +467,376 @@ void testCompilingSingleProjectServer() {
     });
   }
 
+  @Test
+  void testRunTestsProjectServer() {
```
jdneo: Maybe we can consider to separate the test feature tests to a separate file since this is a big one. And I have a feeling that more and more cases will be added for different test framework, and different usage for the same framework.

server/src/main/java/com/microsoft/java/bs/core/internal/services/BuildTargetService.java
```diff
@@ -469,6 +477,86 @@ public ScalacOptionsResult getBuildTargetScalacOptions(ScalacOptionsParams param
...
+            LOGGER.warning("Test params arguments must each be in the form KEY=VALUE. "
+                + testSuites.getEnvironmentVariables());
+            statusCode = StatusCode.ERROR;
+          } else {
+            Map<BuildTargetIdentifier, Map<String, Set<String>>> testClasses = new HashMap<>();
+            for (ScalaTestSuiteSelection testSuiteSelection : testSuites.getSuites()) {
+              Map<String, Set<String>> classesMethods = new HashMap<>();
+              Set<String> methods = new HashSet<>(testSuiteSelection.getTests());
+              classesMethods.put(testSuiteSelection.getClassName(), methods);
+              testClasses.put(params.getTargets().get(0), classesMethods);
```
jdneo: The `classesMethods` is always re-initialized at line 538, makes the testClasses always only contain one test target.

```
Map<String, Set<String>> classesMethods = new HashMap<>();
for (ScalaTestSuiteSelection testSuiteSelection : testSuites.getSuites()) {
    Set<String> methods = new HashSet<>(testSuiteSelection.getTests());
    classesMethods.put(testSuiteSelection.getClassName(), methods);
}
testClasses.put(params.getTargets().get(0), classesMethods);
```

Arthurm1: Yes - it's tricky.  The `TestParams` passes a list of build target ids and a list of test classes.  I've got no way of associating these, i.e. working out which test class applies to which build target.  But luckily Gradle doesn't care - you just supply the classes to test regardless of project.  But we care because `TestReportReporter` has to pass back a build target id.  So I've used the first build target id in all cases.  I'll have a look to see if there's any way to work this out correctly from the Gradle test events but I doubt it.  I assume you don't actually care about build target ids at the BSP client end - you just use the test name data?

server/src/test/java/com/microsoft/java/bs/core/internal/server/BuildTargetServerIntegrationTest.java
Arthurm1: Happy to move the testing to a different integ test.

server/src/main/java/com/microsoft/java/bs/core/internal/services/BuildTargetService.java
jdneo: > I assume you don't actually care about build target ids at the BSP client end - you just use the test name data?

Yes, my client implementation makes sure that only one build target will be included in the request.

jdneo:
BTW, during the integration, I found an issue about the JUnit 5 @Nested support for Gradle Test Launcher: https://github.com/gradle/gradle/issues/29603

Just in case you may have the same problem.

server/src/main/java/com/microsoft/java/bs/core/internal/reporter/TestReportReporter.java
```diff
@@ -0,0 +1,193 @@
...
+            testDuration += result.getEndTime() - result.getStartTime();
+            StatusCode statusCode = StatusCode.OK;
+            TestStatus testStatus = TestStatus.PASSED;
+            String stackTrace = null;
+            if (result instanceof TestFailureResult testFailureResult) {
+              statusCode = StatusCode.ERROR;
+              testStatus = TestStatus.FAILED;
+              stackTrace = testFailureResult.getFailures()
+                  .stream()
+                  .filter(f -> f instanceof DefaultTestAssertionFailure)
```
jdneo: Only extract the stackTrace from DefaultTestAssertionFailure  may not be enough. Take this **junit4** example:

```java
public class ExceptionInBefore {

    @BeforeClass  
    public static void beforeClass() throws Exception {
        throw new Exception("Exception in @BeforeClass");
    };

    @Test
    public void test() {
        
    }
}
```

In this case, the stackTrace is stored in DefaultTestFrameworkFailure

Arthurm1: I've pushed an update to use `TestFailure` instead of `DefaultTestAssertionFailure` and added a test using `@BeforeAll`

Arthurm1:
Added `TestNG` and `Spock` tests.  I've just used whatever is currently sent back for display name etc.  I'm not sure if that's what's wanted but it's a start.

server/src/main/java/com/microsoft/java/bs/core/internal/services/BuildTargetService.java
jdneo: Hi @Arthurm1, what about this? Assume the client will only send one target now?

Arthurm1: @jdneo I've pushed another commit so that the map isn't overwritten

jdneo:
Some test cases failed. Except for that, the PR LGTM. One last thing I hope to do is to separate the related test cases to another file.

I can do that as well. @Arthurm1 Let me know if you want me to do that.

Arthurm1:
@jdneo I can't get the test classes to fail locally.  And I can't see why they would fail in the way they have.  I've added a commit to test out whether there is an issue with messages appearing on listeners from previous tests.  This commit doesn't use the connection cache for tests.  I don't intend for it to be merged - I'm just hoping to see if the tests return a different result.  Could you run the github test actions on this PR again please

jdneo:
I tried to run test on my mac and windows (without the last commit: 
Connector cache test - do not merge). Mac passed. While windows `testSpock()` failed. Haven't got a chance to look into it.

But if it's related with cache. I think we need to separate the test to more small pieces. Like: one test case only to cover one test class/method, which makes it more maintainable.

server/src/main/java/com/microsoft/java/bs/core/internal/gradle/GradleApiConnector.java
```diff
@@ -112,42 +140,100 @@ public GradleSourceSets getGradleSourceSets(URI projectUri, BuildClient client)
...
+                  .setStandardError(errorOut)
+                  .addProgressListener(testReportReporter, OperationType.TEST);
+              if (compileProgressReporter != null) {
+                launcher.addProgressListener(compileProgressReporter, OperationType.TASK);
+              }
+              for (Map.Entry<String, Set<String>> classesMethods : entry.getValue().entrySet()) {
+                if (classesMethods.getValue() != null && !classesMethods.getValue().isEmpty()) {
+                  launcher.withJvmTestMethods(classesMethods.getKey(), classesMethods.getValue());
+                } else {
+                  launcher.withJvmTestClasses(classesMethods.getKey());
```
jdneo: Change to `launcher.withJvmTestClasses(classesMethods.getKey() + "*");` to support the @Nested annotation.

I verified that this will have no side-effect for other kinds of tests

See: https://github.com/gradle/gradle/issues/29603#issuecomment-2190948117

Arthurm1: I've made the change and added a NestedTest into the integ tests.
Can you re-run the github action - I've added some logging to hopefully see what's going on

Arthurm1:
@jdneo I think Github failing is down to the changes in https://github.com/gradle/gradle/pull/28019.

I don't know why Github is picking up that change but locally we are not.  I've updated the tooling API to 8.8 in the latest commit and it should solve the Github issue.

I think there's another issue with Github - I don't think it picks up env vars for some reason.  So I think `testSingleMethodJunit` and `testSingleMethodTestNg` will fail.
Do you know why that would be?  Can we disable those 2 tests for Github?  It's possible that I'm wiping out other env vars that Gradle needs to run - I've had to include `SystemRoot` for Windows.  Maybe more are needed for Github but I don't know which ones.  Gradle seems to only add `SystemRoot`...
https://github.com/gradle/gradle/blob/5cf3ebb43a13d610dd617187498faa61f475df44/platforms/extensibility/test-kit/src/main/java/org/gradle/testkit/runner/internal/DefaultGradleRunner.java#L342

jdneo:
Wait... The GitHub Action runs on ubuntu, I thinking it's nothing todo with the Windows and SystemRoot stuff.

> I don't know why Github is picking up that change but locally we are not. 

Is that because when locally run, the build server picks your gradle local installation(GRADLE_HOME env)? While I believe GitHub does not have that set.

Is that because the gradle version used by `java-test`? If I add the gradle-wrapper.properties file into it, and point to 8.8. The test will pass. Otherwise, the test fails, where gradle complain that getTestDisaplayName() does not exist.

jdneo:
No it should not related with the gradle version, because github pipeline only fails two cases. If it's due to the version, more cases should fail.

But is that safe to call getTestDisaplayName()? Since it's a new API.

Arthurm1:
> But is that safe to call getTestDisplayName()? Since it's a new API.

I'm not sure.  I don't know whether the we can use it because the tooling API is at 8.8 or whether Gradle also needs to be at 8.8.

If I try different versions of Gradle locally I can't get `getTestDisplayName` to fail.
What happens when you get it to fail - is there a `NoSuchMethodError`?

jdneo:
It says: 
```
AbstractMethodError@93 "java.lang.AbstractMethodError: Receiver class org.gradle.internal.build.event.types.DefaultTestDescriptor does not define or inherit an implementation of the resolved method 'abstract java.lang.String getTestDisplayName()' of interface org.gradle.tooling.internal.protocol.events.InternalTestDescriptor."
```

See this commit: https://github.com/microsoft/build-server-for-gradle/pull/144/commits/4c01d1c5115ae3b3934fc28e1fa230208fee40c8, I forced the test project gradle version fix to 8.7 by adding the wrapper property file. Now more cases failed. Which highly like due to the API getTestDisplayName() I guess.

jdneo:
I think maybe we need to:

1. rollback the getTestDisplayName() since it looks like now back compatible.
2. Disable the two failing env tests. I can track it as a debt and solve it in another PR.

Arthurm1:
I've wrapped `getTestDisplayName` so it should handle older versions.

Don't know how to disable tests purely on Github - is there an env var?

Arthurm1:
Added an additional sourceset to test

Arthurm1:
[Here](https://docs.github.com/en/actions/learn-github-actions/variables) is says that "CI=true" will be set for github actions.

I've disabled the 2 env var tests based on this.  Also fixed formatting

jdneo (approved):
LGTM

jdneo:
Thank you @Arthurm1 for your contribution.

We can consider to support test debug delegation. Looks like BSP does not have that flag in request body. Maybe another extended point?

Arthurm1:
@jdneo I think `Metals` used [debugSessionStart](https://build-server-protocol.github.io/docs/specification#debugsessionstart-request) but for `DebugSessionParamsDataKind` used the same data kinds we use for tests e.g. `scala-test-suites-selection` and `scala-test`.

What's kind of interesting is that different clients might want different debug adapters....

`Metals` is primarily for Scala development so it uses https://github.com/scalacenter/scala-debug-adapter
For Kotlin there is https://github.com/fwcd/kotlin-debug-adapter
Java has https://github.com/microsoft/java-debug

I imagine the Java one is good enough for all but it would be handy to be able to specify the type in the BSP command and have the build server kick start the relevant one.  Another thing to propose to add to BSP spec.

jdneo:
Hi @Arthurm1, is the test feature working as expected on your Windows? I tested it today on Windows it's not working unless I remove `launcher.setEnvironmentVariables(envVars);`. But it's fine on Mac.

Arthurm1:
@jdneo Yes it works for me running `gradlew server:test`.

Are you on your `cs/fix-test` branch?  I see you've removed the `SystemRoot` env var setting on that branch and that was needed for Windows to work.

jdneo:
Ah, I mean the code itself. Not test.

I'm testing the integration with the gradle build server test support feature. Everything works fine one Mac. But on my windows, I have to comment `launcher.setEnvironmentVariables(envVars);`, otherwise the tooling api won't run the selected tests successfully.

Not sure if it's related with issue: https://github.com/gradle/gradle/issues/29785. Or it's caused by my own environment.

Arthurm1:
Ah - it works if I add `SystemRoot` property...
```
        // Running Gradle tests on Windows seems to require the SystemRoot env var
        // Otherwise Windows complains "Unrecognized Windows Sockets error: 10106"
        String systemRoot = System.getenv("SystemRoot");
        if (systemRoot != null) {
            env.put("SystemRoot", systemRoot);
        }
```

I should probably add this to `GradleAPIConnector#runTests`.  It's unfortunate that there is no `TestLauncher#addExtraEnvVars` method

Arthurm1:
@jdneo see https://github.com/microsoft/build-server-for-gradle/pull/164

PR #145, opened 2024-05-25
Support buildTarget/run

Arthurm1:
Allows running of `main` methods using `buildTarget/run` BSP request

Supports [`scala-main-class`](https://build-server-protocol.github.io/docs/extensions/scala#scalamainclass) `RunParamsDataKind` as there is no jvm specific data type.

Caveats...
1. StdIn - not connected since `run/readStdin` not supported - need to upgrade BSP to 2.2
2. StdOut - sent as log message since `run/printStdout` not supported - need to upgrade BSP to 2.2
3. StdErr - sent as log message since `run/printStderr` not supported - need to upgrade BSP to 2.2
4. Env vars are not passed -  need to upgrade BSP to 2.2

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,7 @@ Following BSP requests are supported in the current implementation:
 - `buildTarget/javacOptions`
 - `buildTarget/scalacOptions`
 - `buildTarget/test`
+- `buildTarget/run`
 - `workspace/buildTargets`
 - `workspace/reload`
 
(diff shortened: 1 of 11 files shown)


jdneo:
Looks like a lot of features are missing before upgrading to 2.2

Is the code change big for upgrading the BSP to 2.2? If not, maybe we can upgrade first.

Arthurm1:
> Is the code change big for upgrading the BSP to 2.2? If not, maybe we can upgrade first.

I don't think there's much difference.  I'll submit a PR

PR #148, opened 2024-06-06
sidhaarth:bsp-cancellation : BSP Gradle Cancellation

Sidhaarthsr:


jdneo:
Hi @Sidhaarthsr,

There are something that you can make further investigation:

### About the CancelChecker

So, as you can see, In `GradleBuildServer.java` a lot of request handling method has already passed an argument `cc`, which is the cancel checker. Ideally, [when the client sends a cancel request](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#cancelRequest) to the server, we should be able to be notified that via the cancel checker. This capability should be provided by the underlying `lsp4j.jsonrpc` library. I haven't tested it yet, so this is something that you can take a look maybe. If's it's working as expected, than we can let the cancel checker as another parameter for all those request handling method. And check it when necessary. (I'm not sure that check it regularly in another thread is a good idea, do you have any reference for this approach?)

### About cancel the task executed in Gradle Daemon

When we use gradle tooling api to let Gradle daemon do some task for us, Gradle TAPI also supports cancellation.
- https://docs.gradle.org/current/javadoc/org/gradle/tooling/ConfigurableLauncher.html#withCancellationToken(org.gradle.tooling.CancellationToken)
- https://docs.gradle.org/current/javadoc/org/gradle/tooling/GradleConnector.html#newCancellationTokenSource()

Please keep in mind that the Gradle build server and Gradle Daemon are running in two different process. If we just simply cancel the request at Gradle build server side, but not notify the cancellation to the Gradle Daemon. The computation resources are still wasted.

So we need to figure out a way to establish kind of connection between:
- the cancel checker ( makes sure that is working as expected)
- the cancellation token for gradle

jdneo:
@Sidhaarthsr Any issue you've met?

Sidhaarthsr:
@jdneo  The branch got deleted and the PR was closed due to a branch rename. PR #157 is a continuation of this.

PR #122, opened 2023-12-21
Handle composite builds

Arthurm1:
This PR extracts source sets for composite builds - previously, included builds would be ignored.

I've had to get rid of the unique display name code and go with `project [sourceset]` for all target display names.  It just gets far too complicated for little reward when composite builds are added into the mix.

Composite builds have to be handled in the `GradleAPIConnector` as each included build requires a new `ProjectConnection`

PR #154, opened 2024-06-20
Composite Builds using Build Actions

Tanish-Ranjan:
This PR continues the work done by @Arthurm1 for composite-builds on #122, bringing in support for included builds, including dependency substitution using build actions.


Tanish-Ranjan:

Arthurm1:
@Tanish-Ranjan Thanks for the PR.

I see you've used a Gradle `BuildAction`.  Was that so you don't have to query the source sets from `GradleAPIConnector` for each included build?

If that's the best route then you probably don't need the `GradleIncludedBuild` class as the model never returns any.

I think you'll still need to work out the project dependencies across included builds.  Maybe you can put that in the `BuildAction` as well.  I've slightly altered your new test to show what I mean.  This dependency check currently fails...

```java
  @Test
  void testCompositeBuild2() {
    File projectDir = projectPath.resolve("composite-build-2").toFile();
    PreferenceManager preferenceManager = new PreferenceManager();
    preferenceManager.setPreferences(new Preferences());
    GradleApiConnector connector = new GradleApiConnector(preferenceManager);
    GradleSourceSets gradleSourceSets = connector.getGradleSourceSets(projectDir.toURI(), null);
    assertEquals(6, gradleSourceSets.getGradleSourceSets().size());
    GradleSourceSet mainApp = findSourceSet(gradleSourceSets, "app [main]");
    GradleSourceSet mainStringUtils = findSourceSet(gradleSourceSets, "string-utils [main]");
    GradleSourceSet mainNumberUtils = findSourceSet(gradleSourceSets, "number-utils [test]");
    assertHasBuildTargetDependency(mainApp, mainStringUtils);
    assertHasBuildTargetDependency(mainApp, mainNumberUtils);
  }
```

Tanish-Ranjan:
> I see you've used a Gradle `BuildAction`. Was that so you don't have to query the source sets from `GradleAPIConnector` for each included build?

That's correct @Arthurm1. Gradle automatically manages substitution of dependencies. It didn't work previously in [commit f2097d0](https://github.com/microsoft/build-server-for-gradle/pull/154/commits/f2097d003e718ba45116877914d680e50bc41e75) because Gradle was getting scoped to the project, loosing reference of the overall build and hence was not able to perform dependency substitutions from the included builds. With the help of Gradle TAPI's `BuildAction` we are maintaining the build scope while we fetch the `GradleSourceSets`.

Tanish-Ranjan:
> I think you'll still need to work out the project dependencies across included builds.

Thanks for pointing it out. I'll look into it.

server/src/main/java/com/microsoft/java/bs/core/internal/gradle/actions/GetSourceSetsAction.java
```diff
@@ -0,0 +1,65 @@
...
+    Map<String, List<GradleSourceSet>> sourceSets = new HashMap<>();
+    GradleBuild buildModel = buildController.getBuildModel();
+    String rootProjectName = buildModel.getRootProject().getName();
+    fetchModels(buildController, buildModel, sourceSets, rootProjectName);
+    return new DefaultGradleSourceSets(sourceSets.values().stream().flatMap(List::stream)
+        .collect(Collectors.toList()));
+  }
+
+  /**
+   * Fetches source sets from the provided Gradle build model and stores them in a map categorized by project name.
```
reinsch82: did you run `./gradlew build`locally?
this line is too long according to checkStyle rules.

server/src/main/java/com/microsoft/java/bs/core/internal/gradle/actions/GetSourceSetsAction.java
```diff
@@ -0,0 +1,65 @@
...
+public class GetSourceSetsAction implements BuildAction<GradleSourceSets> {
+
+  /**
+   * Executes the build action and retrieves source sets from the Gradle build.
+   *
+   * @return A {@link DefaultGradleSourceSets} object containing all retrieved source sets.
+   */
+  @Override
+  public GradleSourceSets execute(BuildController buildController) {
+    Map<String, List<GradleSourceSet>> sourceSets = new HashMap<>();
```
reinsch82: You could make use of `var`, since it's java 17

```suggestion
    var sourceSets = new HashMap<String, List<GradleSourceSet>>();
```

reinsch82 (approved):
This looks good to me.
Just a few remarks.
I think you have to wait for @jdneo's review though.

server/src/main/java/com/microsoft/java/bs/core/internal/gradle/actions/GetSourceSetsAction.java
Tanish-Ranjan: Fixed ✅

server/src/main/java/com/microsoft/java/bs/core/internal/gradle/actions/GetSourceSetsAction.java
Tanish-Ranjan: Done ✅

server/src/main/java/com/microsoft/java/bs/core/internal/gradle/actions/GetSourceSetsAction.java
```diff
@@ -0,0 +1,110 @@
...
+        rootProjectName);
+
+    // Add dependencies
+    var sourceSets = new ArrayList<GradleSourceSet>();
+    for (var entry : sourceSetToClasspath.entrySet()) {
+
+      var dependencies = new HashSet<BuildTargetDependency>();
+      for (File file : entry.getValue()) {
+        GradleSourceSet otherSourceSet = outputsToSourceSet.get(file);
+        if (otherSourceSet != null) {
```
jdneo: Is it possible that the otherSourceSet is the current sourceset being iterated?

I'm asking because I'm thinking that if it's necessary to check it here.

server/src/main/java/com/microsoft/java/bs/core/internal/gradle/actions/GetSourceSetsAction.java
```diff
@@ -0,0 +1,110 @@
...
+   * @param sourceSetToClasspath A map that associates GradleSourceSet objects with their
+   *                             corresponding classpath files.
+   * @param outputsToSourceSet   A map that associates output files with the GradleSourceSet
+   *                             they belong to.
+   * @param buildName            The name of the root project in the build.
+   */
+  private void fetchModels(
+      BuildController buildController,
+      GradleBuild build,
+      List<String> traversedProjects,
```
jdneo: If it's to avoid cyclic dependencies, HashSet seems better.

plugin/src/main/java/com/microsoft/java/bs/gradle/plugin/SourceSetsModelBuilder.java
```diff
@@ -108,10 +112,13 @@ public Object buildAll(String modelName, Project rootProject) {
         List<File> compileClasspath = new LinkedList<>(sourceSet.getCompileClasspath().getFiles());
```
jdneo, line 112: I'm thinking that if we need to try-catch here. Though the method signature does not declare that it will throw exception.

Will `sourceSet.getCompileClasspath().getFiles()` throw RuntimeException? (maybe a question to @donat @reinsch82 @hegyibalint)

server/src/main/java/com/microsoft/java/bs/core/internal/gradle/actions/GetSourceSetsAction.java
Tanish-Ranjan: I don't think a source set can have a dependency on it's own. Like a source set to output an artifact that exists in it's own compile classpath? It shouldn't be possible I believe. However I would let the Gradle community confirm it.

server/src/main/java/com/microsoft/java/bs/core/internal/gradle/actions/GetSourceSetsAction.java
Tanish-Ranjan: I will wait for other comments to get addressed first. Then I will push a commit resolving all of them.

plugin/src/main/java/com/microsoft/java/bs/gradle/plugin/SourceSetsModelBuilder.java
reinsch82: I don't see anything that would throw there.

jdneo: Before using build action, this line of code throws an exception, see: https://github.com/microsoft/build-server-for-gradle/pull/122#issuecomment-2031099285

So, I'm thinking that wrapping a try-catch for RuntimeException/GradleException

server/src/main/java/com/microsoft/java/bs/core/internal/gradle/actions/GetSourceSetsAction.java
jdneo: At least it do no harm.

plugin/src/main/java/com/microsoft/java/bs/gradle/plugin/SourceSetsModelBuilder.java
Tanish-Ranjan: Yes, it was the case then. [FileCollection#getFiles](https://docs.gradle.org/current/javadoc/org/gradle/api/file/FileCollection.html#getFiles()) used to throw an exception.

server/src/main/java/com/microsoft/java/bs/core/internal/gradle/actions/GetSourceSetsAction.java
Tanish-Ranjan: Okay 👍

model/src/main/java/com/microsoft/java/bs/gradle/model/GradleSourceSetsMetadata.java
```diff
@@ -0,0 +1,15 @@
...
+import java.io.File;
+import java.io.Serializable;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Provides necessary information to build SourceSets.
+ */
+public interface GradleSourceSetsMetadata extends Serializable {
+  Map<GradleSourceSet, List<File>> getGradleSourceSets();
```
jdneo: `getGradleSourceSets` -> `getGradleSourceSetToClasspath` ?

model/src/main/java/com/microsoft/java/bs/gradle/model/GradleSourceSetsMetadata.java
```diff
@@ -0,0 +1,15 @@
+package com.microsoft.java.bs.gradle.model;
+
+import java.io.File;
+import java.io.Serializable;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Provides necessary information to build SourceSets.
```
jdneo: Add more description to explain the purpose of this class?

server/src/main/java/com/microsoft/java/bs/core/internal/gradle/actions/GetSourceSetsAction.java
```diff
@@ -0,0 +1,110 @@
...
+        rootProjectName);
+
+    // Add dependencies
+    var sourceSets = new ArrayList<GradleSourceSet>();
+    for (var entry : sourceSetToClasspath.entrySet()) {
+
+      var dependencies = new HashSet<BuildTargetDependency>();
+      for (File file : entry.getValue()) {
+        GradleSourceSet otherSourceSet = outputsToSourceSet.get(file);
+        if (otherSourceSet != null && entry.getKey() != otherSourceSet) {
```
jdneo: Use `Objects.equals()` instead of `!=`

plugin/src/main/java/com/microsoft/java/bs/gradle/plugin/SourceSetsModelBuilder.java
```diff
@@ -43,11 +42,15 @@
...
   @Override
   public boolean canBuild(String modelName) {
-    return modelName.equals(GradleSourceSets.class.getName());
+    return modelName.equals(GradleSourceSetsMetadata.class.getName());
   }
 
   @Override
   public Object buildAll(String modelName, Project rootProject) {
+
+    Map<GradleSourceSet, List<File>> gradleSourceSets = new HashMap<>();
```
jdneo: rename to `sourceSetToClasspath`?

plugin/src/main/java/com/microsoft/java/bs/gradle/plugin/SourceSetsModelBuilder.java
jdneo: I'll handle it in https://github.com/microsoft/build-server-for-gradle/pull/163

server/src/main/java/com/microsoft/java/bs/core/internal/gradle/actions/GetSourceSetsAction.java
```diff
@@ -0,0 +1,111 @@
...
+   * @param sourceSetToClasspath A map that associates GradleSourceSet objects with their
+   *                             corresponding classpath files.
+   * @param outputsToSourceSet   A map that associates output files with the GradleSourceSet
+   *                             they belong to.
+   * @param buildName            The name of the root project in the build.
+   */
+  private void fetchModels(
+      BuildController buildController,
+      GradleBuild build,
+      HashSet<String> traversedProjects,
```
jdneo: nit: `HashSet<String>` -> `Set<String>`

To align with the below params, since they are Map, not HashMap

jdneo (approved):
LGTM.

Tested with [Gradle official composite build sample](https://docs.gradle.org/current/samples/sample_composite_builds_basics.html) and [a composite build project](https://github.com/microsoft/vscode-gradle/issues/1435) that provided by a user. Both works as expected.

Thank you @Tanish-Ranjan and @Arthurm1 for the contribution, good job!

Tanish-Ranjan:
Big thanks to @Arthurm1 for the initial insights into the tooling API! Your guidance gave this PR a significant head start. Additionally, a huge thanks to all the gsoc mentors (@jdneo @donat @hegyibalint @reinsch82) who provided support throughout this process. Your help was invaluable in getting this completed so quickly.

jdneo:
Ok, so after some investigation, I found that the current solution has some problems:

1. See #169 and #170
2. For the sample project provided in https://github.com/microsoft/vscode-gradle/issues/1435, all the sourceset from two subproject will have the same project path `:`, this is actually not correct, we need to use the build tree path to differentiate them. Otherwise, the build target dependency resolution in `BuildTargetManager` might be wrong.

PR #157, opened 2024-06-24
bsp-cancellation : Support for Gradle Task Cancellation using BSP CancelChecker & TAPI CancellationTokenSource

Sidhaarthsr:
This PR is a continuation of PR [#148](https://github.com/microsoft/build-server-for-gradle/pull/148) which got closed due to branch rename.

### DIFF
--- a/server/src/main/java/com/microsoft/java/bs/core/internal/gradle/GradleApiConnector.java
+++ b/server/src/main/java/com/microsoft/java/bs/core/internal/gradle/GradleApiConnector.java
@@ -65,7 +65,7 @@ public class GradleApiConnector {
    * Get the source sets of the Gradle project.
    *
    * @param projectUri uri of the project
-   * @param client     connection to BSP client
+   * @param client connection to BSP client
    * @return an instance of {@link GradleSourceSets}
    */
   public GradleSourceSets getGradleSourceSets(
@@ -86,7 +86,8 @@ public class GradleApiConnector {
       ModelBuilder<GradleSourceSets> customModelBuilder = Utils.getModelBuilder(
           connection,
           preferenceManager.getPreferences(),
-          GradleSourceSets.class);
+          GradleSourceSets.class
+      );
       customModelBuilder.addProgressListener(reporter,
           OperationType.FILE_DOWNLOAD, OperationType.PROJECT_CONFIGURATION)
           .setStandardError(errorOut)
@@ -97,8 +98,7 @@ public class GradleApiConnector {
       }
       customModelBuilder.addJvmArguments("-Dbsp.gradle.supportedLanguages="
           + String.join(",", preferenceManager.getClientSupportedLanguages()));
-      // since the model returned from Gradle TAPI is a wrapped object, here we
-      // re-construct it
+      // since the model returned from Gradle TAPI is a wrapped object, here we re-construct it
       // via a copy constructor and return as a POJO.
       return new DefaultGradleSourceSets(customModelBuilder.withCancellationToken(token).get());
     } catch (GradleConnectionException | IllegalStateException | IOException e) {
@@ -115,8 +115,8 @@ public class GradleApiConnector {
    * Request Gradle daemon to run the tasks.
    *
    * @param projectUri uri of the project
-   * @param reporter   reporter on feedback from Gradle
-   * @param tasks      tasks to run
+   * @param reporter reporter on feedback from Gradle
+   * @param tasks tasks to run
    */
   public StatusCode runTasks(
       URI projectUri,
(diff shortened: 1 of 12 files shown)


jdneo:
@Sidhaarthsr Please go through the CLA according to https://github.com/microsoft/build-server-for-gradle/pull/157#issuecomment-2186398514

server/src/main/java/com/microsoft/java/bs/core/internal/managers/BuildTargetManager.java
```diff
@@ -68,8 +68,8 @@ public List<BuildTargetIdentifier> store(GradleSourceSets gradleSourceSets) {
             false /* canTest */,
             false /* canRun */,
             false /* canDebug */
-          )
-      );
+        )
+    );
```
jdneo: In this PR, there are some changes due to the format change like here. I suggest to revert changes like this to limit the size of the PR.

// Sorry that the repo does not have a format profile right now, we can add some later as an engineering work.

Sidhaarthsr:

server/src/main/java/com/microsoft/java/bs/core/internal/managers/BuildTargetManager.java
Sidhaarthsr: Hi @jdneo , I tried to take care of it in the previous commit but the accidental Shift+Alt+F presses brought them back. Reverted them back again.

Sidhaarthsr:
> @Sidhaarthsr Please go through the CLA according to [#157 (comment)](https://github.com/microsoft/build-server-for-gradle/pull/157#issuecomment-2186398514)

Done

server/src/main/java/com/microsoft/java/bs/core/internal/gradle/GradleApiConnector.java
```diff
@@ -65,7 +68,13 @@ public String getGradleVersion(URI projectUri) {
    * @param client connection to BSP client
    * @return an instance of {@link GradleSourceSets}
    */
-  public GradleSourceSets getGradleSourceSets(URI projectUri, BuildClient client) {
+  public GradleSourceSets getGradleSourceSets(
+      URI projectUri,
+      BuildClient client,
+      final CancelChecker cancelChecker,
+      final CancellationToken token) throws CancellationException {
+    cancelChecker.checkCanceled();
```
jdneo, line 76: It looks not the right way to do the cancel check.

My understanding about this change is:

We checked the cancel status at the entry of the method, and may have two result

- cancelChecker.checkCanceled() throws CancellationException. In such case, the below code will not be executed, which is fine.
- Nothing happens because of the reuqest is not cancelled. Then code will continue to be executed. But, please be aware that, in this case, no one will call `tokenSource.cancel();`. If the request is cancelled after this point, the gradle will still run its tasks without cancellation. In other words, the `.withCancellationToken(token)` does not take any effect.

Sidhaarthsr: Actually we're checking cancellation in first line for all the stages and we're using the same CancellationTokenSource for all Gradle Operations. So if cancellation was requested after checking cancelChecker.checkCanceled() in the current stage it'll be caught in the subsequent stage and when we call tokenSource.cancel() all the Gradle Operations involved in the prior stage will be cancelled.

This can only be an issue when this happens in the last stage. So to guard this, maybe can check for cancellation in the start and end as well for all stages.

jdneo: Sorry I'm little bit confused here. Could you help me understand more about this example?

Let's say the code passes line 76. And execute to line 103: `customModelBuilder.withCancellationToken(token).get().`. In such case, how will the customModelBuilder get be cancelled?

Sidhaarthsr: public void reloadWorkspace(final CancelChecker cancelChecker,
      final CancellationToken token) throws CancellationException {
    cancelChecker.checkCanceled();

    List<BuildTargetIdentifier> changedTargets = updateBuildTargets(cancelChecker, token);
    if (!changedTargets.isEmpty()) {
      notifyBuildTargetsChanged(changedTargets, cancelChecker);
    }
  }

  For instance, if we consider the above use case from which getGradleSourceSets method is called and if the cancellation was issued during or after line 76, the cancellation will still be caught in the notifyBuildTargetsChanged method which executes after this and when tokenSource.cancel() is called from the GradleBuildServer class in the catch block, all the operations associated with the token passed will be cancelled.

  It's a best-effort cancellation

jdneo: I doubt that...

Since you are creating new token source for every request/notification. They are all independent. Which means that, cancellation triggered from reload won't affect others because they are different tokens.

jdneo: We need to find a way to persist the relationship between each protocol cancel checker and the gradle cancel token.

For example, you can have a kind of registry holding a map to store the relationship. Meanwhile, a separate thread will keep check the cancel status for each of the cancel checker. If one is cancelled, then call `cancel()` for the token. And you may also pay attention to maintain the map cache, all the out-of-date entries need to be removed.

This is a rough idea for your reference, but there might be solutions better than this...

Arthurm1:
@Sidhaarthsr Are you still working on this?

Sidhaarthsr:
No, @Arthurm1

Arthurm1:
@Sidhaarthsr Has anyone taken it on?  I'd be happy with finishing it off

Sidhaarthsr:
Sure @Arthurm1! Thanks

jdneo:
See: https://github.com/microsoft/build-server-for-gradle/pull/179

Thank you @Sidhaarthsr for your contribution!

PR #160, opened 2024-07-01
LanguageExtension downcast resolution

Tanish-Ranjan:
This PR removes the usage of a class loader for downcasting `LanguageExtesnion` to `JavaExtension` and `ScalaExtension`. Previously, the client JVM would attempt to determine the appropriate sub-interface type using a class loader.

### Changes:

- Introduced new methods in the `LanguageExtension` interface:
  - `isJavaExtension()` and `isScalaExtension()`: Checks if the current implementation is a specific sub-interface type.
  - `getAsJavaExtension()` and `getAsScalaExtension()`: Returns the instance as the specific sub-interface type if applicable.
- Moved the conversion logic from the client JVM to the Gradle daemon JVM, where the actual implementation of the `LanguageExtension` interface resides. This ensures correct type information is available for downcasting.
- Updated respective test cases to reflect the new approach.

### DIFF
--- a/model/build.gradle
+++ b/model/build.gradle
@@ -3,5 +3,5 @@ java {
 }
 
 dependencies {
-  implementation 'org.gradle:gradle-tooling-api:8.7'
+  implementation 'org.gradle:gradle-tooling-api:8.8'
 }
--- a/model/build.gradle
+++ b/model/build.gradle
@@ -1,7 +1,3 @@
 java {
   sourceCompatibility = JavaVersion.VERSION_1_8
 }
-
-dependencies {
-  implementation 'org.gradle:gradle-tooling-api:8.8'
-}
--- a/model/build.gradle
+++ b/model/build.gradle
@@ -1,3 +1,7 @@
 java {
   sourceCompatibility = JavaVersion.VERSION_1_8
 }
+
+dependencies {
+  implementation 'org.gradle:gradle-tooling-api:8.7'
+}
--- a/model/src/main/java/com/microsoft/java/bs/gradle/model/LanguageExtension.java
+++ b/model/src/main/java/com/microsoft/java/bs/gradle/model/LanguageExtension.java
@@ -3,25 +3,52 @@
 
 package com.microsoft.java.bs.gradle.model;
 
-import org.gradle.internal.impldep.javax.annotation.Nullable;
-
 import java.io.Serializable;
 
 /**
- * parent interface for language extensions.
+ * Interface representing a language extension.
+ *
+ * @see JavaExtension
+ * @see ScalaExtension
  */
 public interface LanguageExtension extends Serializable {
 
-  @Nullable
+  /**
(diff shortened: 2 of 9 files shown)


model/build.gradle
```diff
@@ -1,3 +1,7 @@
 java {
   sourceCompatibility = JavaVersion.VERSION_1_8
 }
+
+dependencies {
+  implementation 'org.gradle:gradle-tooling-api:8.7'
```
jdneo: Let's use 8.8, which is the latest

model/src/main/java/com/microsoft/java/bs/gradle/model/LanguageExtension.java
```diff
@@ -3,17 +3,25 @@
 
 package com.microsoft.java.bs.gradle.model;
 
+import org.gradle.internal.impldep.javax.annotation.Nullable;
```
jdneo: Hey, @hegyibalint do you know if this is fine here? Because I see the Nullable class comes from gradle internal package.

jdneo (commented):
LGTM. Just some minor issues.

model/build.gradle
Tanish-Ranjan: Updated 👍

model/src/main/java/com/microsoft/java/bs/gradle/model/LanguageExtension.java
hegyibalint: I would rather not, as I assume IDE/QA tooling wouldn't pick up this re-packaged class.
If you want to have the annotations, you can depend on it directly just like this:

https://github.com/gradle/gradle/blob/d40c6121608cad99eb818fa698a4daa3eac99aee/build-logic/dependency-modules/src/main/kotlin/gradlebuild/modules/extension/ExternalModulesExtension.kt#L108

jdneo: I see, thanks @hegyibalint 

@Tanish-Ranjan I think this can be another separate engineer work to enable null check to this project. For now, I think we can just remove those annotation in this PR, and make it clear that it may return null in the javadoc.

Tanish-Ranjan: Done ✅

model/build.gradle
```diff
@@ -1,3 +1,7 @@
 java {
   sourceCompatibility = JavaVersion.VERSION_1_8
 }
+
+dependencies {
+  implementation 'org.gradle:gradle-tooling-api:8.8'
+}
```
Arthurm1: Is this needed now `@Nullable` is gone?

jdneo: Yes it's not needed. I'm just about to append a commit to remove it

jdneo (approved):

jdneo:
Thank you @Tanish-Ranjan for your contribution! You can continue to work on #154 now.

PR #164, opened 2024-07-03
Set env vars for Windows testing

Arthurm1:
Set `SystemRoot` in env vars when running tests.  Required to run Gradle tests in Windows

### DIFF
--- a/server/src/main/java/com/microsoft/java/bs/core/internal/gradle/GradleApiConnector.java
+++ b/server/src/main/java/com/microsoft/java/bs/core/internal/gradle/GradleApiConnector.java
@@ -197,7 +197,14 @@ public class GradleApiConnector {
             launcher.withArguments(args);
             launcher.setJvmArguments(jvmOptions);
             // env vars requires Gradle >= 3.5
-            launcher.setEnvironmentVariables(envVars);
+            if (envVars != null) {
+              // Running Gradle tests on Windows seems to require the `SystemRoot` env var
+              // Otherwise Windows complains "Unrecognized Windows Sockets error: 10106"
+              // Assumption is that current env vars plus specified env vars are all wanted.
+              Map<String, String> allEnvVars = new HashMap<>(System.getenv());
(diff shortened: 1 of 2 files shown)


server/src/main/java/com/microsoft/java/bs/core/internal/gradle/GradleApiConnector.java
```diff
@@ -197,7 +197,16 @@ public StatusCode runTests(URI projectUri,
             launcher.withArguments(args);
             launcher.setJvmArguments(jvmOptions);
             // env vars requires Gradle >= 3.5
-            launcher.setEnvironmentVariables(envVars);
+            if (envVars != null) {
+              // Running Gradle tests on Windows seems to require the SystemRoot env var
+              // Otherwise Windows complains "Unrecognized Windows Sockets error: 10106"
+              // Would check Windows using org.gradle.internal.os.OperatingSystem but it's internal
+              String systemRoot = System.getenv("SystemRoot");
```
jdneo: Does that mean, we should first get the existing envs, and then append the customized ones?

Arthurm1: I don't know.  Maybe.  Would be nice if the BSP spec and Gradle had both `envVars` and `additionalEnvVars`.
I can change it to set all env vars if you want.

jdneo: Maybe we can have a try:

1. Get existing envs and append customized ones
2. Remove the disable flag for those two env tests
3. See if this can fix the issue :)

Arthurm1:
@jdneo can you run the github tests - I've included all env vars and removed the `CI` test disable

jdneo:
Yes 👍 

I'll do some more tests in the client integration tomorrow. And merge it if everything is fine.

jdneo (approved):

jdneo:
Thank you @Arthurm1!

Issue #76, opened 2023-08-22
Let client show error messages when incompatible issue comes


jdneo:
See: https://github.com/microsoft/build-server-for-gradle/pull/74#discussion_r1301168154

jdneo:
Addressed by #165

Issue #75, opened 2023-08-22
Give a retry when user specified Gradle Java home is incompatible


jdneo:
See: https://github.com/microsoft/build-server-for-gradle/pull/74#discussion_r1301168154

jdneo:
Addressed by https://github.com/microsoft/build-server-for-gradle/pull/165

PR #165, opened 2024-07-04
Find compatible GradleJavaHome and notify client in case of incompatibility

Tanish-Ranjan:
Addressing [issue 75](https://github.com/microsoft/build-server-for-gradle/issues/75) and [issue 76](https://github.com/microsoft/build-server-for-gradle/issues/76)


server/src/main/java/com/microsoft/java/bs/core/internal/services/LifecycleService.java
```diff
@@ -155,6 +164,13 @@ private void updateGradleJavaHomeIfNecessary(URI rootUri) {
         File jdkInstallation = getJdkToLaunchDaemon(preferences.getJdks(), highestJavaVersion);
         if (jdkInstallation != null) {
           preferences.setGradleJavaHome(jdkInstallation.getAbsolutePath());
+        } else {
+          new ClientNotifier(client).sendNotification(
```
jdneo: Is it possible to expand the compatibility check scope?

Some cases for consideration:

1. If user specifies Gradle Java Home in preference, is it a valid Java home? Is it compatible with the gradle version being used?
2. If user does not specify the gradle version, we still can infer the gradle version used for the project, can we do the compatibility check?

Tanish-Ranjan: Commit [8731678](https://github.com/microsoft/build-server-for-gradle/pull/165/commits/87316784aaf454d061a832bec0a38694301f8149) address this. We are now checking for `GradleJavaHome`, if it is available and compatible. If not we move to find a compatible JDK from the list of JDKs we get from preferences.

Tanish-Ranjan:
To check if the selected JDK is compatible or not we are now comparing the feature release of the Java Version. Previously, if we retrieved the latest compatible JDK version of say `21` and the client has JDK version `21.0.1`. Then `Runtime.Version#compareTo` it would say the client JDK version is greater than the highest compatible Java version. So I have resorted to using the feature release for compatibility check.

server/src/main/java/com/microsoft/java/bs/core/internal/gradle/GradleApiConnector.java
```diff
@@ -51,23 +51,40 @@ public GradleApiConnector(PreferenceManager preferenceManager) {
...
 
-  private String getGradleVersion(ProjectConnection connection) {
-    BuildEnvironment model = connection
+  /**
+   * Extracts the GradleJavaHome for the give project.
+   *
+   * @param projectUri URI of the project to get the gradle java home for.
+   * @return GradleJavaHome of the project or null upon failure.
+   */
+  public File getGradleJavaHome(URI projectUri) {
```
jdneo: This method will return the Java home which used to launch build server itself, do I understand correctly?

Tanish-Ranjan: No, I don't think so. This method gets `GradleJavaHome` for the given gradle project from the [BuildEnvironment](https://docs.gradle.org/current/javadoc/org/gradle/tooling/model/build/BuildEnvironment.html) > [JavaEnvironment](https://docs.gradle.org/current/javadoc/org/gradle/tooling/model/build/JavaEnvironment.html). 

Gradle gets the `GradleJavaHome` property in the following [priority order](https://docs.gradle.org/current/userguide/build_environment.html#priority_for_configurations):
1. Command Line arg
2. Gradle Properties File
3. Environment Variable

In our case only the last two will be valid so, this method will return the `GradleJavaHome` defined in the `gradle.properties` file, otherwise it will fallback to using the `JAVA_HOME` environment variable for the JDK location.

server/src/main/java/com/microsoft/java/bs/core/internal/services/LifecycleService.java
```diff
@@ -119,61 +134,114 @@ public void exit() {
...
+
+    String gradleVersion;
+    if (buildKind == GradleBuildKind.SPECIFIED_VERSION) {
+      gradleVersion = preferences.getGradleVersion();
+    } else {
+      gradleVersion = connector.getGradleVersion(rootUri);
+    }
+
+    if (StringUtils.isEmpty(gradleVersion)) {
+      return null;
```
jdneo: Do we know when `gradleVersion` will be null?

server/src/main/java/com/microsoft/java/bs/core/internal/utils/JavaUtils.java
```diff
@@ -0,0 +1,79 @@
...
+  /**
+   * Reads the Java Version from the given JDK file.
+   *
+   * @param jdkFile JDK file to read the java version from.
+   * @return Java Version of the given JDK or null upon failure.
+   */
+  public static String getJavaVersionFromFile(File jdkFile) throws IOException {
+
+    ProcessBuilder processBuilder =
+        new ProcessBuilder(jdkFile.getAbsolutePath() + "/bin/java", "-version");
```
jdneo, line 28: what will happen if the executable file does not exist?

server/src/main/java/com/microsoft/java/bs/core/internal/services/LifecycleService.java
```diff
@@ -119,61 +134,114 @@ public void exit() {
...
+            preferences.getJdks(),
+            oldestCompatibleVersion,
+            latestCompatibleVersion
+        );
       }
+
     }
+
+    // Notify client, for no compatible JDK can be found
+    new ClientNotifier(client).sendNotification(
```
jdneo: I'm thinking that if this is over design at this stage for the `ClientNotifier`? Does that mean we need to create a new notifier instance every time we want to send some msg to the client?

I like the idea of wrap things like taskid & originId into it. Maybe at current stage, just call the method from `client` directly? In the future, if we found that this is not enough, we can refactor to introduce sth like the notifier. WDYT?

server/src/main/java/com/microsoft/java/bs/core/internal/gradle/GradleApiConnector.java
jdneo: Thank you, I learnt from it!

Just a small suggestion: would you like to do some verification about the behavior? Because I found that some time, the behavior from command line is different from calling the TAPI. Just in case...

Tanish-Ranjan: Okay. I will verify it.

server/src/main/java/com/microsoft/java/bs/core/internal/services/LifecycleService.java
Tanish-Ranjan: I think you meant empty? The answer is yes, when retrieving the Gradle version using `GradleApiConnector#getGradleVersion` from the `BuildEnvironment` and a `BuildException` gets thrown. In this case we will get an empty string.

server/src/main/java/com/microsoft/java/bs/core/internal/utils/JavaUtils.java
Tanish-Ranjan: That should ideally throw an `IOException` when we try to start the process using `processBuilder.start()`. If the exception occurs it will be up to the class calling this utility function to handle it.

I think we can have more unit tests for this method, right now we are only simulating the error in `testGetJavaVersionFromFileForException`.

server/src/main/java/com/microsoft/java/bs/core/internal/services/LifecycleService.java
Tanish-Ranjan: If there is no requirement for the `taskId` and `originId` at this stage, I could make the `sendNotification` method static and there won't be the need to create new instances of ClientNotifier.

jdneo: I would prefer directly using `client`.

server/src/main/java/com/microsoft/java/bs/core/internal/utils/JavaUtils.java
jdneo: Ok, let's add one for file not exist case

server/src/main/java/com/microsoft/java/bs/core/internal/services/LifecycleService.java
jdneo: Do you know when will `BuildEnvironment` get thrown? Just asking for curiosity.

Tanish-Ranjan: The check already existed in the codebase. Although [BuildExceptions](https://docs.gradle.org/current/javadoc/org/gradle/tooling/BuildException.html) are thrown when a model cannot be built. Based on [this](https://stackoverflow.com/questions/53991804/could-not-fetch-model-of-type-buildenvironment-using-gradle-distribution) I would say building the `BuildEnvironment` model would throw an exception due to incorrectly defined Gradle properties.

server/src/main/java/com/microsoft/java/bs/core/internal/utils/JavaUtils.java
Tanish-Ranjan: Added in commit [1e1ad04](https://github.com/microsoft/build-server-for-gradle/pull/165/commits/1e1ad040873483729109c47bc7f489b24f0afa67)

server/src/main/java/com/microsoft/java/bs/core/internal/services/LifecycleService.java
Tanish-Ranjan: Lifecycle service directly uses the client to send notifications with [847c8a2](https://github.com/microsoft/build-server-for-gradle/pull/165/commits/847c8a25e0d409e03f2628beea5aa87f27c8dbd2)

server/src/main/java/com/microsoft/java/bs/core/internal/gradle/GradleApiConnector.java
```diff
@@ -51,23 +51,40 @@ public GradleApiConnector(PreferenceManager preferenceManager) {
...
+    } catch (BuildException e) {
+      LOGGER.severe("Failed to get Java version: " + e.getMessage());
+      return null;
+    }
+  }
+
+  private BuildEnvironment getBuildEnvironment(ProjectConnection connection) {
+    return connection
         .model(BuildEnvironment.class)
         .withArguments("--no-daemon")
```
jdneo: `--no-daemon` does not have effect when using tooling api

server/src/main/java/com/microsoft/java/bs/core/internal/gradle/GradleApiConnector.java
```diff
@@ -153,18 +170,20 @@ public StatusCode runTasks(URI projectUri, ProgressReporter reporter, String...
...
       BuildClient client, String originId,
-      CompileProgressReporter compileProgressReporter) {
+      CompileProgressReporter compileProgressReporter
+  ) {
 
     StatusCode statusCode = StatusCode.OK;
     ProgressReporter reporter = new DefaultProgressReporter(client);
     try (ProjectConnection connection = getGradleConnector(projectUri).connect()) {
-      String gradleVersion = getGradleVersion(connection);
+      String gradleVersion = getGradleVersion(projectUri);
```
jdneo: Could you reuse the connection here? For example, overloading the `getGradleVersion()`?

server/src/main/java/com/microsoft/java/bs/core/internal/services/LifecycleService.java
```diff
@@ -119,61 +134,117 @@ public void exit() {
...
 
   /**
-   * Try to update the Gradle Java home if:
-   * <ul>
-   *   <li>Gradle Java home is not set.</li>
-   *   <li>A valid JDK can be found to launch Gradle.</li>
-   * </ul>
-   *
-   * <p>The JDK installation path string will be set to {@link Preferences#gradleJavaHome}.
+   * Finds a suitable version of JDK to use for compilation of Java programs.
```
jdneo: Let's say a JDK compatible with the target Gradle version. Because it's the jdk which is used for Gradle operations (for example running tasks or acquiring model information). It's not only for compilation.

server/src/main/java/com/microsoft/java/bs/core/internal/services/LifecycleService.java
```diff
@@ -119,61 +134,117 @@ public void exit() {
...
 
-      if (StringUtils.isNotBlank(gradleVersion)) {
-        map = TelemetryUtils.getMetadataMap("gradleVersion", gradleVersion);
-        LOGGER.log(Level.INFO, "Gradle version: " + gradleVersion, map);
-        String highestJavaVersion = Utils.getHighestCompatibleJavaVersion(gradleVersion);
-        File jdkInstallation = getJdkToLaunchDaemon(preferences.getJdks(), highestJavaVersion);
-        if (jdkInstallation != null) {
-          preferences.setGradleJavaHome(jdkInstallation.getAbsolutePath());
+      // Use GradleJavaHome if compatible
+      if (gradleJavaHome != null) {
```
jdneo: Will `gradleJavaHome` null here? Given we have logic at line 152

server/src/main/java/com/microsoft/java/bs/core/internal/gradle/GradleApiConnector.java
Tanish-Ranjan: Yes, I saw it get merged in #172. When this PR is ready for merge I will merge the last commit from the develop branch into this, resolving any merge conflicts. This should get fixed there.

server/src/main/java/com/microsoft/java/bs/core/internal/services/LifecycleService.java
Tanish-Ranjan: Yes, it will be null if the `GradleApiConnector#getGradleJavaHome` gets a `BuildException`. In this case we will get a null value.

Tanish-Ranjan:
The JDK lookup for gradle seems to be different from the way it is listed in the [docs](https://docs.gradle.org/current/userguide/build_environment.html#priority_for_configurations).

Priority order I found out:
1. Command Line args
2. Gradle Properties
3. Installed JDK (gets from the registry I think)
4. JAVA_HOME (Env variable)

server/src/main/java/com/microsoft/java/bs/core/internal/gradle/GradleApiConnector.java
Tanish-Ranjan: Continued [here](https://github.com/microsoft/build-server-for-gradle/pull/165#issuecomment-2232644129).

server/src/main/java/com/microsoft/java/bs/core/internal/services/LifecycleService.java
Tanish-Ranjan: Addressed in commit [fbf550b](https://github.com/microsoft/build-server-for-gradle/pull/165/commits/fbf550b288559f5f92a474307885ce56e417d962)

server/src/main/java/com/microsoft/java/bs/core/internal/gradle/GradleApiConnector.java
Tanish-Ranjan: Addressed in commit [fbf550b](https://github.com/microsoft/build-server-for-gradle/pull/165/commits/fbf550b288559f5f92a474307885ce56e417d962)

server/src/main/java/com/microsoft/java/bs/core/internal/services/LifecycleService.java
jdneo: Ok, my next question is that will `GradleApiConnector#getGradleJavaHome` gurantee that the found java home is compatible?

server/src/main/java/com/microsoft/java/bs/core/internal/gradle/GradleApiConnector.java
jdneo: Or you can just rebase the PR based on the develop branch

server/src/main/java/com/microsoft/java/bs/core/internal/services/LifecycleService.java
Tanish-Ranjan: No. It is only going to return the `GradleJavaHome`, looking for it in the priority order listed in [gradle docs](https://docs.gradle.org/current/userguide/build_environment.html#priority_for_configurations). It only queries for a `GradleJavaHome` and does not check for compatibility. It can even return non existent filepath if the user has wrong configurations for the project (eg. non existent jdk path in gradle.properties or JAVA_HOME).

server/src/main/java/com/microsoft/java/bs/core/internal/services/LifecycleService.java
```diff
@@ -119,61 +134,117 @@ public void exit() {
...
+          }
+        } catch (IOException e) {
+          LOGGER.severe("Invalid GradleJavaHome: " + e.getMessage());
         }
+
+      }
+
+      // Pick a compatible JDK from the JDKs available in Preferences
+      if (preferences.getJdks() != null && preferences.getJdks().isEmpty()) {
+        return getJdkToLaunchDaemon(
```
jdneo: If the preference has set some jdks but no one is compatible with gradle, since we return here, the client will not be notified.

Tanish-Ranjan: Addressed in commit [c7f5d83](https://github.com/microsoft/build-server-for-gradle/pull/165/commits/c7f5d838683fa585efa8d5ffb25ebe9b87d112ea)

server/src/main/java/com/microsoft/java/bs/core/internal/services/LifecycleService.java
```diff
@@ -119,61 +134,120 @@ public void exit() {
...
+      String latestCompatibleJavaVersion
+  ) {
 
     Entry<String, String> selected = null;
     for (Entry<String, String> jdk : jdks.entrySet()) {
       String javaVersion = jdk.getKey();
-      if (Version.parse(javaVersion).compareTo(Version.parse(highestJavaVersion)) <= 0
-          && (selected == null || Version.parse(selected.getKey())
-              .compareTo(Version.parse(javaVersion)) < 0)) {
+      Boolean isSelectedVersionHigher =
```
jdneo: nit:

1. There is no need to use object Boolean here, just primitive `boolean` should be fine.
2. Seems that name the variable to `isHigherThanSelected` makes more sense.

server/src/main/java/com/microsoft/java/bs/core/internal/services/LifecycleService.java
```diff
@@ -119,61 +134,120 @@ public void exit() {
...
+    String latestCompatibleVersion = Utils.getLatestCompatibleJavaVersion(gradleVersion);
+    String oldestCompatibleVersion = Utils.getOldestCompatibleJavaVersion();
+
+    if (StringUtils.isNotBlank(latestCompatibleVersion)) {
+
+      // Use GradleJavaHome if compatible
+      if (gradleJavaHome != null) {
+
+        try {
+          String gradleJavaHomeVersion = JavaUtils.getJavaVersionFromFile(gradleJavaHome);
```
jdneo: What will happen if, `JavaUtils.getJavaVersionFromFile` returns a string which is not a purely string? Will it throw exception in the `JavaUtils.isCompatible()`? Will the exception be caught?

Tanish-Ranjan: What do you mean by purely string? `JavaUtils#getJavaVersionFromFile` returns a string containing the version of the jdk or `null` upon failure. I will add a null check to prevent `NullPointerException` in `JavaUtils#isCompatible`.

jdneo: Sorry, my typo. 

I mean, for example, assume `JavaUtils.getJavaVersionFromFile` returns "foo17.0.1"

Tanish-Ranjan: Is that a special case in which the version string contains additional characters?
For example if we run the command `java -version`, the supposed output would be in the given format correct?
```
java -version
java version "21.0.3" 2024-04-16 LTS
```
And now we are extracting the version printed between the double quotes.

jdneo: I'm not sure. Since you know there are so many different java distribution, I'm not sure if they will all print the version like that.

I mean. no need to cover all those cases. If we found that the version string cannot be correctly parsed, just makes sure the server will not fail down due to the unhandled exceptions.

server/src/main/java/com/microsoft/java/bs/core/internal/services/LifecycleService.java
Tanish-Ranjan: Addressed in commit [e7de6d8](https://github.com/microsoft/build-server-for-gradle/pull/165/commits/e7de6d870c1893f46e4d25f1bfb2b59ffed35218)

server/src/main/java/com/microsoft/java/bs/core/internal/services/LifecycleService.java
Tanish-Ranjan: Addressed in commit [e7de6d8](https://github.com/microsoft/build-server-for-gradle/pull/165/commits/e7de6d870c1893f46e4d25f1bfb2b59ffed35218)

server/src/main/java/com/microsoft/java/bs/core/internal/services/LifecycleService.java
```diff
@@ -119,61 +134,121 @@ public void exit() {
...
+
     Preferences preferences = preferenceManager.getPreferences();
-    if (preferences.getJdks() == null || preferences.getJdks().isEmpty()) {
-      return;
+    String filePath = preferences.getGradleJavaHome();
+    File gradleJavaHome;
+    if (StringUtils.isNotEmpty(filePath)) {
+      gradleJavaHome = new File(filePath);
+    } else {
+      gradleJavaHome = connector.getGradleJavaHome(rootUri);
```
jdneo: Some new thoughts here:

One assumption: According to my past experience, most of the users will use default value, and the default value for the gradle java home and gradle version are empty.

That means, the code here will call tooling api to get the `BuildEnvironment` twice. This can be improved by just get the `BuildEnvironment` once, and later, get the gradle and java version on demand.

server/src/main/java/com/microsoft/java/bs/core/internal/utils/JavaUtils.java
```diff
@@ -0,0 +1,79 @@
...
+  /**
+   * Checks if the given Java Version is compatible based on feature release.
+   *
+   * @param javaVersion Java version to check for compatibility.
+   * @param oldestJavaVersion oldest compatible Java version.
+   * @param latestJavaVersion latest compatible Java version.
+   * @return true if given javaVersion is less than or equal to the latestJavaVersion
+   *        and greater than or equal to the oldestJavaVersion.
+   */
+  public static Boolean isCompatible(
```
jdneo: nit: Use `boolean`

server/src/main/java/com/microsoft/java/bs/core/internal/utils/JavaUtils.java
```diff
@@ -0,0 +1,79 @@
...
+      String javaVersion,
+      String oldestJavaVersion,
+      String latestJavaVersion
+  ) {
+
+    Version versionToCheck = Version.parse(javaVersion);
```
jdneo, line 74: catch the `IllegalArgumentException` and `NumberFormatException`

consider to add null check as well

server/src/main/java/com/microsoft/java/bs/core/internal/utils/JavaUtils.java
```diff
@@ -0,0 +1,79 @@
...
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.lang.Runtime.Version;
+
+/**
+ * Utility class for Java.
+ */
+public class JavaUtils {
+
+  public JavaUtils() {
```
jdneo: should be private?

server/src/test/java/com/microsoft/java/bs/core/internal/server/BuildTargetServerIntegrationTest.java
```diff
@@ -477,6 +535,87 @@ void testCompilingSingleProjectServer() {
     });
   }
 
+  @Test
+  void testIncompatibleUserJavaHomeProjectServer() {
```
jdneo: Makes more sense to put it into `LifecycleServiceIntegrationTests`?

server/src/main/java/com/microsoft/java/bs/core/internal/services/LifecycleService.java
Tanish-Ranjan: Okay. We can save the last queried `BuildEnvironment` and use that if needed again. Even though the `BuildEnvironment` is project dependent, we are only using it for the root project. I will mention the use case in the javaDoc for future references.

server/src/main/java/com/microsoft/java/bs/core/internal/utils/JavaUtils.java
Tanish-Ranjan: I didn't add those checks there because it is inside a utility class, and there are often instances where the caller would like to handle the exception on its own. So we may just throw the error there and handle the exception at the call site.

server/src/test/java/com/microsoft/java/bs/core/internal/server/BuildTargetServerIntegrationTest.java
Tanish-Ranjan: Are you proposing to create a new test file? I didn't find any existing test file called `LifecycleServiceIntegrationTests`.

server/src/main/java/com/microsoft/java/bs/core/internal/services/LifecycleService.java
jdneo: Do you mean cache the `BuildEnvironment` in the connector?

server/src/main/java/com/microsoft/java/bs/core/internal/utils/JavaUtils.java
jdneo: Ok, just make sure there is no uncaught exception.

server/src/test/java/com/microsoft/java/bs/core/internal/server/BuildTargetServerIntegrationTest.java
jdneo: Yes it is.

server/src/main/java/com/microsoft/java/bs/core/internal/services/LifecycleService.java
Tanish-Ranjan: Yes in the `GradleApiConnector#getBuildEnvironment` we can cache the `BuildEnvironment` obtained and then return it instead for the next calls. We may add a parameter to this method as well to force it to fetch a new `BuildEnvironment` if needed.

jdneo: Might be a little bit over-design. Just get it once here is good enough

server/src/main/java/com/microsoft/java/bs/core/internal/utils/JavaUtils.java
Tanish-Ranjan: Fixed in commit [04ed182](https://github.com/microsoft/build-server-for-gradle/pull/165/commits/04ed182c48bf04e766b4189900cdfbb470550b50)

server/src/main/java/com/microsoft/java/bs/core/internal/utils/JavaUtils.java
Tanish-Ranjan: Fixed in commit [04ed182](https://github.com/microsoft/build-server-for-gradle/pull/165/commits/04ed182c48bf04e766b4189900cdfbb470550b50)

server/src/main/java/com/microsoft/java/bs/core/internal/utils/JavaUtils.java
Tanish-Ranjan: Fixed in commit [04ed182](https://github.com/microsoft/build-server-for-gradle/pull/165/commits/04ed182c48bf04e766b4189900cdfbb470550b50)

server/src/main/java/com/microsoft/java/bs/core/internal/services/LifecycleService.java
Tanish-Ranjan: Fixed in commit [04ed182](https://github.com/microsoft/build-server-for-gradle/pull/165/commits/04ed182c48bf04e766b4189900cdfbb470550b50)

server/src/test/java/com/microsoft/java/bs/core/internal/server/BuildTargetServerIntegrationTest.java
Tanish-Ranjan: Fixed in commit https://github.com/microsoft/build-server-for-gradle/pull/165/commits/8ed01255f35de1f7ab772357cfc30e44a3f26174

jdneo:
Hi @Tanish-Ranjan, The PR has some conflicts with the develop branch now. Could you please resolve them?

server/src/main/java/com/microsoft/java/bs/core/internal/gradle/GradleApiConnector.java
```diff
@@ -53,22 +53,63 @@ public GradleApiConnector(PreferenceManager preferenceManager) {
...
+
+  /**
+   * Extracts the BuildEnvironment model for the given project.
+   *
+   * @param projectUri URI of the project ot get the gradle java home for.
+   * @return BuildEnvironment of the project or {@code null} upon failure.
+   */
+  public BuildEnvironment getBuildEnvironment(URI projectUri) {
+    try (ProjectConnection connection = getGradleConnector(projectUri).connect()) {
+      return connection
```
jdneo: return getBuildEnvironment(connection);

server/src/test/java/com/microsoft/java/bs/core/internal/server/IntegrationTest.java
```diff
@@ -0,0 +1,431 @@
...
+        new BuildClientCapabilities(SupportedLanguages.allBspNames);
+    return new InitializeBuildParams(
+        "test-client",
+        "0.1.0",
+        "0.1.0",
+        root.toURI().toString(),
+        capabilities);
+  }
+
+  protected static InitializeBuildParams getInitializedBuildParamsWithJdks(
```
jdneo: Put this as a member of LifecycleServiceIntegrationTest?

server/src/test/java/com/microsoft/java/bs/core/internal/server/IntegrationTest.java
```diff
@@ -0,0 +1,431 @@
...
+          && Objects.equals(testName.getMethodName(), methodName)
+          && Objects.equals(getTestNameHierarchy(testName), testNames);
+    }
+
+    private static String testNameAsString(TestName testName) {
+      return testName.getSuiteName() + "," + testName.getClassName() + ","
+          + testName.getMethodName() + "," + getTestNameHierarchy(testName);
+    }
+
+    TestStartEx getTestStart(String suiteName, String className, String methodName,
```
jdneo: Since this is the base class, I think it should only provide the minimum required members. These test related method better move to the `BuildTargetServiceIntegrationTest`?

server/src/main/java/com/microsoft/java/bs/core/internal/gradle/GradleApiConnector.java
Tanish-Ranjan: Fixed in https://github.com/microsoft/build-server-for-gradle/pull/165/commits/7e4749a704b01421c5b6282ea0204d8856b32636

server/src/test/java/com/microsoft/java/bs/core/internal/server/IntegrationTest.java
Tanish-Ranjan: Fixed in https://github.com/microsoft/build-server-for-gradle/pull/165/commits/7e4749a704b01421c5b6282ea0204d8856b32636

server/src/test/java/com/microsoft/java/bs/core/internal/server/IntegrationTest.java
Tanish-Ranjan: Fixed in https://github.com/microsoft/build-server-for-gradle/pull/165/commits/7e4749a704b01421c5b6282ea0204d8856b32636

Tanish-Ranjan: Fixed in https://github.com/microsoft/build-server-for-gradle/pull/165/commits/7e4749a704b01421c5b6282ea0204d8856b32636

server/src/test/java/com/microsoft/java/bs/core/internal/server/LifecycleServiceIntegrationTest.java
```diff
@@ -0,0 +1,137 @@
...
+
+    BuildClientCapabilities capabilities =
+        new BuildClientCapabilities(SupportedLanguages.allBspNames);
+    final InitializeBuildParams initParams = new InitializeBuildParams(
+        "test-client",
+        "0.1.0",
+        "0.1.0",
+        root.toURI().toString(),
+        capabilities
+    );
```
jdneo: Looks like we can directly get the initParams by `IntegrationTest.getInitializeBuildParams(projectDir)`, which can remove duplicated code

Tanish-Ranjan: Fixed in https://github.com/microsoft/build-server-for-gradle/pull/165/commits/24fdf827bf77e943ea93778a3b3fdd7b78a9b94c

Tanish-Ranjan:
Prioritizing Gradle default behavior for java home detection with https://github.com/microsoft/build-server-for-gradle/pull/165/commits/3b6888733339c8f42229b18fa9cecfa09f84d500

reinsch82 (approved):
Looks good to me.

thx

jdneo (approved):
LGTM

PR #168, opened 2024-07-09
Parallelize source set retrieval.

Arthurm1:
This PR got away from me slightly so has a number of changes.

The idea here was to retrieve sourcesets concurrently by creating an action for each project and then submitting all those actions at once.  If Gradle > 6.7 then it will attempt to execute these actions in parallel.
Scala in particular can be slow to populate some of its information so this is faster.  I haven't tested to see if it speeds up any large public Gradle projects like `Gradle` or `Kafka`.
This also simplifies the `SourceSetsModelBuilder` as the data retrieved is purely for 1 project.

Changes...
1. Changed `SourceSetsModelBuilder` to only populate 1 project to allow parallel retrieval in `GradleSourceSetsAction`.
2. Changed `GetSourceSetsAction` to get the model for all projects/subprojects/included builds.
3. Allow multiple output dirs per sourceSet - the single output dir wasn't working when projects had Scala & Java
4. Shift output dir population to specific language model builders - output dir is language dependent
5. Map archives to multiple output dirs for correct project dependency population - wasn't working out sourceset dependencies correctly.
6. Use `GradleBuild#getEditableBuilds` instead of `GradleBuild#getIncludedBuilds` if available.  This will include the `buildSrc` project. See [here](https://docs.gradle.org/current/javadoc/org/gradle/tooling/model/gradle/GradleBuild.html#getEditableBuilds())
7. Substitute project jars on the classpath with their original classes dirs.  In `Scala` and in included builds, the classpath includes the project jars rather than the classes directories.  So for BSP clients, they now realize that the class files are from project sources and can navigate to the source rather than decompiling the classfile from the jar.
8. Removed `main` source set only limitation on inter-project dependencies.  Any project with custom configs (or one using `TestFixtures`) needs this.
9. Upgraded Spock dependency in tests so tests run on JDK 21.
10. Remove `GradleSourceSet#ProjectName` - it's not used
11. Added a test with a missing repository to check for exceptions thrown by `getCompileClasspath`.  Although this really means that the user's build file is wrong but at least they don't get an exception from the plugin.


plugin/src/main/java/com/microsoft/java/bs/gradle/plugin/SourceSetsModelBuilder.java
```diff
@@ -42,198 +44,209 @@
...
+    gradleSourceSet.setExtensions(extensions);
+    gradleSourceSet.setSourceOutputDirs(sourceOutputDirs);
 
-        sourceSetsToClasspath.put(gradleSourceSet, compileClasspath);
+    // classpath
+    try {
+      Set<File> files = sourceSet.getCompileClasspath().getFiles();
+      List<File> compileClasspath = new LinkedList<>(files);
+      gradleSourceSet.setCompileClasspath(compileClasspath);
+    } catch (ResolveException ex) {
```
jdneo: Use `GradleException`? Since it may throw `org.gradle.internal.resolve.ModuleVersionNotFoundException` as well

https://github.com/microsoft/build-server-for-gradle/pull/122#issuecomment-2031099285

Arthurm1: I think it varies depending on the Gradle version you use.  I added the `missing-repository` test and found that `ResolveException` is thrown in `SourceSetsModelBuilder` for some versions and `ModuleVersionNotFoundException` is thrown in `DependencyCollector` in others.  So I've added `GradleException` to the `DependencyCollector` code.
I can change `ResolveException` to `GradleException` here if you want but I can't reproduce that.

jdneo: I'd like to use GradleException. Missing repository is one case to make the invocation failed. But there must be some other cases that we don't know which will fail the invocation too. Using GradleException makes me feel more confident

Arthurm1: Done

Arthurm1:
@jdneo  I've pushed all the language specific stuff to be stored in the `LanguageExtension` (e.g. source dirs, generated dirs) instead of queried from the `LanguageModelBuilder`.
This means that there are fewer lookups of the compilation task and so an increase in performance.

I've also moved `SupportedLanguages` out of `GradleBuildServerPlugin`.  It was a `static` but it was populated in the class constructor and so kept getting overwritten when running in parallel causing exceptions.

I've tested anecdotally using [Kafka](https://github.com/apache/kafka).  I'm getting speed ups on sourceset retrieval of around 5x when `org.gradle.parallel=true` compared to before this PR.

Arthurm1:
I've experienced `ConcurrentModificationException` on `project.getTasks().withType(XX)` so added `synchronized` to guard against them.

I've tested against [Gradle](https://github.com/gradle/gradle) and it's 12x faster to retrieve source sets on my machine.

Arthurm1:
Fixed test.  The supported languages weren't being supplied in all `GradleBuildServerPluginTest` tests.

`GradleBuildServerPluginTest` now uses `GetSourceSetsAction` instead of retrieving the model directly, so all Gradle versions are now tested.

jdneo:
Hi @Arthurm1, I'm working on some other stuff recently. I'll make sure to review this PR once I've finished the stuff.

model/src/main/java/com/microsoft/java/bs/gradle/model/actions/GetSourceSetsAction.java
```diff
@@ -37,79 +37,144 @@ public class GetSourceSetsAction implements BuildAction<GradleSourceSets> {
...
     }
 
-    return new DefaultGradleSourceSets(sourceSets);
+    // since the model returned from Gradle TAPI is a wrapped object, here we re-construct it
+    // via a copy constructorso we can treat as a DefaultGradleSourceSet and
+    // populate source set dependencies.
+    List<GradleSourceSet> sourceSets = buildController.run(projectActions).stream()
+        .flatMap(ss -> ss.getGradleSourceSets().stream())
+        .map(ss -> new DefaultGradleSourceSet(ss))
+        .collect(Collectors.toList());
```
Tanish-Ranjan, line 105: Do we still need a copy constructor here @jdneo?

jdneo: Haven't found a time to do the verification. Filed an issue to track: https://github.com/microsoft/build-server-for-gradle/issues/175

model/src/main/java/com/microsoft/java/bs/gradle/model/GradleSourceSet.java
```diff
@@ -24,11 +24,6 @@ public interface GradleSourceSet extends Serializable {
    */
   public String getDisplayName();
 
-  /**
-   * Equivalent to {@code org.gradle.api.Project.getName()}. 
-   */
-  public String getProjectName();
```
jdneo: The project name is used by JDT.LS when creating a new `IProject` instance during project import. Let's preserve it.

plugin/src/main/java/com/microsoft/java/bs/gradle/plugin/dependency/DependencyCollector.java
```diff
@@ -42,70 +45,72 @@ public class DependencyCollector {
...
   /**
    * Resolve and collect dependencies from a {@link SourceSet}.
    */
-  public void collectByConfigurationNames(Set<String> configurationNames) {
-    if (GradleVersion.current().compareTo(GradleVersion.version("3.3")) < 0) {
-      List<ResolvedConfiguration> configs = project.getConfigurations().stream()
-              .filter(configuration -> configurationNames.contains(configuration.getName()))
+  public static Set<GradleModuleDependency> getModuleDependencies(Project project,
+      Set<String> configurationNames) {
+    if (GradleVersion.current().compareTo(GradleVersion.version("4.0")) < 0) {
```
jdneo, line 53: Is this due to the `ArtifactView.ViewConfiguration` compatibility?

Arthurm1: Yes - when I added this in the previous PR I put `3.3` as the version when Gradle started to support it but actually it should be `4.0`.
The code paths haven't really changed in `DependencyCollector`, I've just switched it from populating the dependencies into the source set to returning a list of dependencies.
Then the exclusions are removed later in the build action.

model/src/main/java/com/microsoft/java/bs/gradle/model/GradleSourceSet.java
Arthurm1: I've reinstated project name

model/src/main/java/com/microsoft/java/bs/gradle/model/JavaExtension.java
```diff
@@ -35,4 +35,9 @@ public interface JavaExtension extends LanguageExtension {
    * The list of compiler arguments.
    */
   public List<String> getCompilerArgs();
+
+  /**
+   * The classes directory.
+   */
+  File getClassesDir();
```
jdneo: The `LanguageExtension` already declares the method `File getClassesDir();`. Is this on purpose? Same to ScalaExtension

model/src/main/java/com/microsoft/java/bs/gradle/model/actions/GetSourceSetsAction.java
```diff
@@ -37,79 +37,144 @@ public class GetSourceSetsAction implements BuildAction<GradleSourceSets> {
...
+  /**
+   * Fetches source sets from the provided Gradle build model.
+   *
+   * @param buildController The Gradle build controller used to interact with the build.
+   * @param builds The Gradle build models representing the build and included builds.
+   */
+  private List<GradleSourceSet> fetchModels(BuildController buildController,
+      Collection<GradleBuild> builds) {
 
+    List<GetSourceSetAction> projectActions = new ArrayList<GetSourceSetAction>();
```
jdneo: nit: new ArrayList<>();

Arthurm1: Fixed

model/src/main/java/com/microsoft/java/bs/gradle/model/JavaExtension.java
Arthurm1: Fixed

jdneo:
Hi @Arthurm1, should I add `org.gradle.parallel` in `gradle.properties` when verifying this PR? (I guess the answer is `No`, right?)

jdneo:
I can observe the perf boost for this change. Opening the repo `Gradle`, the request for `workspace/buildTarget` is 3 times faster than before on my machine.

Meanwhile, I encountered an error with this change: https://github.com/microsoft/vscode-gradle/issues/1566. This error cannot be observed in the develop branch. Looks like we need to add the dedup logic in the project dependency resolution?

Arthurm1:
> Hi @Arthurm1, should I add `org.gradle.parallel` in `gradle.properties` when verifying this PR? (I guess the answer is `No`, right?)

I think using `org.gradle.parallel` or enabling the configuration cache should cause the model to be built in parallel.  But it also depends on which gradle version you're using.  I've been trying both.

Arthurm1:
> Meanwhile, I encountered an error with this change: [microsoft/vscode-gradle#1566](https://github.com/microsoft/vscode-gradle/issues/1566). This error cannot be observed in the develop branch. Looks like we need to add the dedup logic in the project dependency resolution?

I can't see why this would happen.

The classpaths for `daemon-main [crossVersionTest]` and `daemon-main [testFixtures]` both contain...
```
file:/C:/repodir/gradle/platforms/core-execution/snapshots/build/classes/java/main/
file:/C:/repodir/gradle/platforms/core-execution/snapshots/build/classes/groovy/main
file:/C:/repodir/gradle/platforms/core-execution/snapshots/build/resources/main
file:/C:/repodir/gradle/platforms/core-execution/snapshots/build/generated-resources/gradle-snapshots-classpath/
file:/C:/repodir/gradle/platforms/core-execution/snapshots/build/classes/java/testFixtures
file:/C:/repodir/gradle/platforms/core-execution/snapshots/build/classes/groovy/testFixtures/
file:/C:/repodir/gradle/platforms/core-execution/snapshots/build/resources/testFixtures
```
and `daemon-main [crossVersionTest]` contains dependencies...
```
file:/C:/repodir/gradle/platforms/core-execution/snapshots/?sourceset\u003dtestFixtures
file:/C:/repodir/gradle/platforms/core-execution/snapshots/?sourceset\u003dmain
```

So it doesn't look like there are duplicates unless the `testFixtures` and `crossVersionTest` source sets are being merged into a single `daemon-main` project?

It doesn't appear to be duplicating display names either.

jdneo:
> So it doesn't look like there are duplicates unless the testFixtures and crossVersionTest source sets are being merged into a single daemon-main project

Ah I see. This might be the reason. Because the JDT project model does not have equivalent concept for source set in Gradle. In JDT, one project can only have one classpath. So they are merged together.

Sorry for the confusion.

jdneo:
> I think using org.gradle.parallel or enabling the configuration cache should cause the model to be built in parallel. But it also depends on which gradle version you're using. I've been trying both.

Question about this: I tried this change without setting `org.gradle.parallel`, and observe the perf boost as well. Is this required?

Arthurm1:
> Ah I see. This might be the reason. Because the JDT project model does not have equivalent concept for source set in Gradle. In JDT, one project can only have one classpath. So they are merged together.

Do you special case the `test` source set then?  If not, I can't see how it wouldn't always fail.

Arthurm1:
> Question about this: I tried this change without setting `org.gradle.parallel`, and observe the perf boost as well. Is this required?

I don't know for sure.  I think you either have to set `org.gradle.parallel=true` or `org.gradle.configuration-cache=true`

jdneo:
> Do you special case the test source set then? If not, I can't see how it wouldn't always fail.

Yes, in JDT, each classpath entry can have an attribute named `test`, but that's all.

Arthurm1:
> Yes, in JDT, each classpath entry can have an attribute named `test`, but that's all.

Ah - does the gradle project fail without this PR then?

jdneo:
> Ah - does the gradle project fail without this PR then?

No, it doesn't

server/src/main/java/com/microsoft/java/bs/core/internal/managers/BuildTargetManager.java
```diff
@@ -82,13 +83,11 @@ public List<BuildTargetIdentifier> store(GradleSourceSets gradleSourceSets) {
         changedTargets.add(btId);
       }
       newCache.put(btId, buildTarget);
-      // Store the relationship between the project dir and the build target id.
-      // 'test' and other source sets are ignored.
-      if ("main".equals(sourceSet.getSourceSetName())) {
-        projectPathToBuildTargetId.put(sourceSet.getProjectDir().getAbsolutePath(), btId);
-      }
```
Arthurm1, line 89: @jdneo The issue with project dependencies exception might be down to me removing this.  Now all sourceset dependencies are returned which I think is correct.
E.g. `IntegTests` might depend on `testFixtures` and `main`

jdneo: Ok, never mind. I've already add dedup logic at client side: https://github.com/microsoft/vscode-gradle/pull/1568

This is more likely a client side issue due to the limited capability of client.

jdneo (approved):
LGTM. Thank you @Arthurm1 for your contribution

PR #170, opened 2024-07-10
fix - Unsupported class file for the build action class

### LINKED ISSUE #169: The compile of build action
We introduced the build action to solve composite build in #154. There is a problem that the build action class is in the server project, which is compiled by jdk 17. But in real, the build action class runs inside the gradle daemon, so if the gradle daemon is not running on jdk >= 17, error like `Unsupported class file major version 61` will be thrown.

Looks like we should move `GetSourceSetsAction` into the `model` module.

I guess this is something we need to do before the parallel work @Arthurm1. I'll raise a PR soon.

jdneo:
- Move the GetSourceSetsAction class to the model module, to ensure that the GetSourceSetsAction class is compiled on level 1.8.
- Set the required properties to the BuildActionExecuter.

fix #169

### DIFF
--- a/model/build.gradle
+++ b/model/build.gradle
@@ -1,3 +1,11 @@
+plugins {
+    id 'java-library'
+}
+
 java {
   sourceCompatibility = JavaVersion.VERSION_1_8
+}
+
+dependencies {
+    api 'org.gradle:gradle-tooling-api:8.8'
 }
--- a/model/src/main/java/com/microsoft/java/bs/gradle/model/actions/GetSourceSetsAction.java
+++ b/model/src/main/java/com/microsoft/java/bs/gradle/model/actions/GetSourceSetsAction.java
@@ -1,4 +1,7 @@
-package com.microsoft.java.bs.core.internal.gradle.actions;
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+package com.microsoft.java.bs.gradle.model.actions;
 
 import com.microsoft.java.bs.gradle.model.BuildTargetDependency;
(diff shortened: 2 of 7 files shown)


jdneo:
_// cc @Tanish-Ranjan, @Arthurm1_

Arthurm1:
LGTM - makes sense to have the action in the `Model` project.  I'd possibly go further and set the java version for `server` and `plugin` to 8 as well since the 17 features are hardly used and then more JDK versions are supported, but it's not for this PR.

jdneo:
>  I'd possibly go further and set the java version for server and plugin to 8 as well

the plugin is already 8 now: https://github.com/microsoft/build-server-for-gradle/blob/develop/plugin/build.gradle#L19. Server is now on 17. Yes making it on 8 will expand the usage. But some refactoring needs to be done. I can do that in another PR.

Tanish-Ranjan:
I have a different thought on this. The `Model` module only contains model classes which are used by `Plugin` and `Server`. It made the most sense for me to have the BuildAction in the `Server` functionality vise, although as you pointed out, we have a slight catch, that the BuildActions are executed within the Gradle Daemon.

So, I would add this to changes:
- renaming the `Model` module to `Internal` (Something being used internally by both `Plugin` and `Server`) as it is no longer dedicated for models only.

jdneo:
@Tanish-Ranjan We can put it a debt. I'll discuss this with Donat to see what's the best answer.

For now, maybe we can merge this change first and release an updat. (I observed this problem because I saw the import success rate drop down from the BI)

testforstephen (approved):

PR #171, opened 2024-07-10
fix - Use project dir instead of project path

jdneo:
- In composite build scenario, the project path is not based on the build tree root. Besides, the .getBuildTreePath() is not available before Gradle 8.3, we use the project dir to identify a Gradle project instead. The project dir is used in dependency resolution.

Fix the point 2 in https://github.com/microsoft/build-server-for-gradle/pull/154#issuecomment-2219647109

### DIFF
--- a/model/src/main/java/com/microsoft/java/bs/gradle/model/BuildTargetDependency.java
+++ b/model/src/main/java/com/microsoft/java/bs/gradle/model/BuildTargetDependency.java
@@ -10,7 +10,7 @@ import java.io.Serializable;
  * In Java this is a project:sourceset dependency on a project:sourceset.
  */
 public interface BuildTargetDependency extends Serializable {
-  public String getProjectPath();
+  public String getProjectDir();
 
   public String getSourceSetName();
 }
--- a/model/src/main/java/com/microsoft/java/bs/gradle/model/impl/DefaultBuildTargetDependency.java
+++ b/model/src/main/java/com/microsoft/java/bs/gradle/model/impl/DefaultBuildTargetDependency.java
@@ -14,19 +14,18 @@ import com.microsoft.java.bs.gradle.model.GradleSourceSet;
 public class DefaultBuildTargetDependency implements BuildTargetDependency {
   private static final long serialVersionUID = 1L;
 
-  private String projectPath;
+  private String projectDir;
 
   private String sourceSetName;
 
-  public DefaultBuildTargetDependency(String projectPath, String sourceSetName) {
-    this.projectPath = projectPath;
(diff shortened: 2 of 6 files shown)


jdneo:
// cc @Tanish-Ranjan, @Arthurm1

Arthurm1:
So `project.getPath()` is not unique when using included builds?  We might need to also look at what goes into `gradleSourceSet.setDisplayName`.  It might be better to do in the action after full retrieval.  Maybe set display name to full path and then pare it down to differences in the build action?
I did something similar [here](https://github.com/scalacenter/gradle-bloop/blob/506dc933d0148abe6bfe88f477db9e9d1b0fdf07/src/main/scala/bloop/integrations/gradle/model/BloopConverter.scala#L825).  I can look to reproduce in another PR?

jdneo:
> So project.getPath() is not unique when using included builds?

No they are not. In the sample project that I referenced, both project path is `:`.

> It might be better to do in the action after full retrieval. Maybe set display name to full path and then pare it down to differences in the build action?
I did something similar [here](https://github.com/scalacenter/gradle-bloop/blob/506dc933d0148abe6bfe88f477db9e9d1b0fdf07/src/main/scala/bloop/integrations/gradle/model/BloopConverter.scala#L825). I can look to reproduce in another PR?

Sure, feel free to address it in another PR if you want :)

Tanish-Ranjan:
> So project.getPath() is not unique when using included builds?

It won't. Any information that comes from the `project` will be stripped down to the [rootProject](https://github.com/microsoft/build-server-for-gradle/blob/3b691da501a9b9f7a2f0df84e726ba4187725573/plugin/src/main/java/com/microsoft/java/bs/gradle/plugin/SourceSetsModelBuilder.java#L49). `SourceSetsModelBuilder` considers the project it receives in the `buildAll` parameter, to be the root project. This is why we had to resort to `BuildActions` in #154, as it contains the complete information of the project hierarchy. It queries for each sub project in the build. This is where the `SourceSetsModelBuilder` starts building the requested model unaware of the entire project hierarchy and thinks the sub project we are querying is actually the root project.

Now because of the way model builders work we cannot rely on hierarchical information from it, such as project path.

Hope I explained it well. Please ask any questions if you have queries.

jdneo:
Another solution is to get the build tree path by casting the `Project` to `DefaultProject`, and call `getIdentityPath()`. But that is an internal API, and since the project dir works fine for now, we can use project dir currently.

Just to share this information in case we need to revisit this problem in the future.

testforstephen (approved):

PR #173, opened 2024-07-16
feat - Support resolving debug & release variant from Android Java project

Tanish-Ranjan:
This PR resolves android projects to build GradleSourceSets out of build variants.


jdneo:
Hi @Tanish-Ranjan, I saw there are a lot of android resource files in the test project. Are they all necessary for testing purpose?

Tanish-Ranjan:
No. The project was generated by the Android Studio, so there are a lot of resources generated along with it. I will narrow it down as I add the unit tests.

plugin/src/main/java/com/microsoft/java/bs/gradle/plugin/SourceSetsModelBuilder.java
```diff
@@ -37,6 +39,7 @@
 import com.microsoft.java.bs.gradle.model.impl.DefaultGradleSourceSet;
 import com.microsoft.java.bs.gradle.model.impl.DefaultGradleSourceSets;
 import com.microsoft.java.bs.gradle.plugin.dependency.DependencyCollector;
+import org.jetbrains.annotations.NotNull;
```
jdneo: `org.jetbrains.annotations` will be unavailable in other IDEs.

Tanish-Ranjan: Addressed in https://github.com/microsoft/build-server-for-gradle/pull/173/commits/9120f6841af25c6b698e754f67c36a136ca6cb35

testProjects/android-test/mylibrary/consumer-rules.pro
jdneo, line 1: Is this empty file required?

plugin/src/main/java/com/microsoft/java/bs/gradle/plugin/utils/AndroidUtils.java
```diff
@@ -0,0 +1,404 @@
...
+        }
+      }
+      // add R.jar file
+      String taskName = "process" + capitalize(variantName) + "Resources";
+      Task processResourcesTask = project.getTasks().findByName(taskName);
+      if (processResourcesTask != null) {
+        Object output = processResourcesTask.getClass()
+            .getMethod("getRClassOutputJar").invoke(processResourcesTask);
+        RegularFile file = (RegularFile) output.getClass()
+            .getMethod("get").invoke(output);
```
jdneo: How about wrap the reflections to a utility method?

testProjects/android-test/mylibrary/consumer-rules.pro
Tanish-Ranjan: Removed default proguard files in https://github.com/microsoft/build-server-for-gradle/pull/173/commits/e4cc0c244f988508817e6e77796159e9bd0a61d9

plugin/src/main/java/com/microsoft/java/bs/gradle/plugin/utils/AndroidUtils.java
Tanish-Ranjan: Addressed in https://github.com/microsoft/build-server-for-gradle/pull/173/commits/e4cc0c244f988508817e6e77796159e9bd0a61d9

testProjects/android-test/app/src/main/res/mipmap/ic_launcher_round.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<adaptive-icon xmlns:android="http://schemas.android.com/apk/res/android">
+    <background android:drawable="@drawable/ic_launcher_background" />
```
jdneo: If we do not reference `@drawable/ic_launcher_xxx`, can those be removed as well?

Tanish-Ranjan: I guess we can remove all the resource files as we are not querying those in any unit tests.

Tanish-Ranjan: Removed in https://github.com/microsoft/build-server-for-gradle/pull/173/commits/04ee668f23c8e64397b647fe3d63a92636eba3ef

jdneo: Ok, cool!

Tanish-Ranjan:
`testAndroidSourceSets` can be ignored. I left a note in the test. The two missing dependencies are from Android SDK which will be added only if `ANDROID_HOME` is set.

Tanish-Ranjan:
There a couple of limitations/improvements possible for Android support:

1. **Intermediate generated sources**
   Android generates a lot of intermediate sources such as AIDL, Render script, C/CPP, etc. Unfortunately AGP doesn't provide any APIs to retrieve these files. `R.jar` is one of these intermediate sources which is crucial for Android development so as a workaround I have retrieved the file using the process task for the build variant however, in some cases the task may not be registered during the sync in which case the dependency for `R.jar` is not provided.
   
   I have opened a ticket here: https://github.com/microsoft/build-server-for-gradle/issues/181
   
2. **Other project type support**
   There various kinds of projects that come under Android development such as Application, Library, Instant App, Dynamic Modules, Android Test, etc. I have added support for the most commonly used projects - Application and Library but the current implementation may require further enhancements to support other kinds of projects.
   
   I have opened a ticket here: https://github.com/microsoft/build-server-for-gradle/issues/182
   
3. **Test variants**
   My implementation takes into account any user defined build variants and the default build variants - `debug` and `release` except the default test variants - `test` and `androidTest`.

   I have opened a ticket here: https://github.com/microsoft/build-server-for-gradle/issues/183

4. **`ANDROID_HOME` fallback logic**
   Android Components to be used by an Android project is configured via `ANDROID_HOME` environment variable. If the property doesn't exist then we are not providing the dependency. This implementation can be improved via fallback logic similar to what we did for `JAVA_HOME` in #165.
   
   I have opened a ticket here: https://github.com/microsoft/build-server-for-gradle/issues/184

jdneo:
Hi @Tanish-Ranjan,

Thank you for listing the follow-ups. Could you please copy the context information to the related issues as well. Currently, the linked issues are all empty.

plugin/src/main/java/com/microsoft/java/bs/gradle/plugin/utils/AndroidUtils.java
```diff
@@ -263,6 +278,29 @@ private static GradleSourceSet convertVariantToGradleSourceSet(Project project,
...
+        extension.setJavaHome(DefaultInstalledJdk.current().getJavaHome());
+        extension.setJavaVersion(DefaultInstalledJdk.current().getJavaVersion().getMajorVersion());
+
+        extension.setSourceDirs(sourceDirs);
+        extension.setGeneratedSourceDirs(generatedSources);
+        extension.setClassesDir(sourceOutputs.iterator().next());
+        extension.setCompileTaskName(compileTaskName);
+
+        extension.setCompilerArgs(compilerArgs);
+        extension.setSourceCompatibility(getSourceCompatibility(compilerArgs));
```
jdneo, line 288: I think that what we really need is the source compatibility and target compatibility.

Things like the java home and java version may not required since the android.jar already contains the types from jdk.

And things like sourceDirs, generatedSources, etc... are not required either because you already set them to the source set.

Tanish-Ranjan: Addressed in https://github.com/microsoft/build-server-for-gradle/pull/173/commits/f129e626874179194621bca70bd645af966d2fb2

Tanish-Ranjan: I tried to integrate these changes with the vscode-gradle extension, and the external library mapping fails on the extension side.

jdneo: The vscode-gradle needs to be updated accordingly.

1. https://github.com/microsoft/vscode-gradle/blob/07878470c8dcc8a61b153671c78357d04743fa10/extension/jdtls.ext/com.microsoft.gradle.bs.importer/src/com/microsoft/gradle/bs/importer/GradleBuildServerBuildSupport.java#L587-L590

Here, instead of throw an exception, we should allow the empty java home and gradle version. Just leave log that they are empty.

2. Here https://github.com/microsoft/vscode-gradle/blob/07878470c8dcc8a61b153671c78357d04743fa10/extension/jdtls.ext/com.microsoft.gradle.bs.importer/src/com/microsoft/gradle/bs/importer/GradleBuildServerBuildSupport.java#L516-L540. When the Java Home is empty, the code here is unnecessary

Tanish-Ranjan: Ah I see.

Question though: If the sources, and generated sources are optional since we are already passing it in the build target info, why do we add it to extensions in the first place? Shouldn't it be the same for other project type support (java and scala)?

jdneo: Because each language can have its own source/generated sources, that's the part of language feature in Gradle: https://github.com/microsoft/build-server-for-gradle/blob/395b97757da8d0d82bc2c68c4aca4331bea4e584/plugin/src/main/java/com/microsoft/java/bs/gradle/plugin/SourceSetsModelBuilder.java#L100-L102

Tanish-Ranjan: Updated vscode-gradle with https://github.com/microsoft/vscode-gradle/pull/1594/commits/9021f77659658c560a8da5905e1d65e7719e964b

jdneo (approved):
LGTM

Tanish-Ranjan:
With default build types `debug` and `release`, this PR also resolved any other user configured build types. It's the default test variants - `unitTest` and `androidTest` which it failed to extract.

PR #178, opened 2024-08-01
Test gradle.properties jvm args are used

Arthurm1:
Adds a test case to check that jvmargs in `gradle.properties` are respected.

It's possible to set jvm args at various points of the tooling api.  This checks they're being set in the `ProjectConnection`.  If the tooling API spins off another process to retrieve the model and that doesn't use the same jvm args, then that wouldn't be tested for here.

Uses [BuildEnvironment](https://docs.gradle.org/current/javadoc/org/gradle/tooling/model/build/BuildEnvironment.html) to check jvm args.

### DIFF
--- a/server/src/main/java/com/microsoft/java/bs/core/internal/gradle/GradleApiConnector.java
+++ b/server/src/main/java/com/microsoft/java/bs/core/internal/gradle/GradleApiConnector.java
@@ -56,8 +56,8 @@ public class GradleApiConnector {
    * Get the Gradle version of the project.
    */
   public String getGradleVersion(URI projectUri) {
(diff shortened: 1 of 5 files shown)


jdneo:
So it turns out the jvm args are picked by gradle?

jdneo:
@timreebow Could you share more information? Or is there any other way that we can leverage for testing?

Arthurm1:
> So it turns out the jvm args are picked by gradle?

It seems so.  https://github.com/gradle/gradle/issues/22856 seems to back that up it but also suggests that there is a priority order so I did wonder if they were being overwritten somewhere rather than not being used.

Issue #175, opened 2024-07-22
Is copy constructor needed indeed?


jdneo:
See: https://github.com/microsoft/build-server-for-gradle/blob/6aa16d8ee6d7afefcf88d4e58704360c57825b4e/server/src/main/java/com/microsoft/java/bs/core/internal/gradle/GradleApiConnector.java#L103-L105

jdneo:
See: https://github.com/microsoft/build-server-for-gradle/pull/188#issuecomment-2313920729

Looks like, if we want to get rid of the copy constructor, we need to add an api for the model, somthing like `getGradleSourceSetInstance()`. Similar to what we did in the LanguageExtension: https://github.com/microsoft/build-server-for-gradle/blob/develop/model/src/main/java/com/microsoft/java/bs/gradle/model/LanguageExtension.java#L70

PR #188, opened 2024-08-27
Handle java source/target defined at extension level

### LINKED ISSUE #3721: 'Type Pattern' being reported as an error even though Java source level is above 16
While modding for the game Mindustry, I noticed that I get the error `The Java feature 'Type Patterns' is only available with source level 16 and above - Java(4195411)` while doing 'Type Pattern' even though the `sourceLevel` is set to 17 (see linked mod template repository down below).
The explanation for this error might be because `targetLevel` is set to 8 and the `--release 8` argument is also added. If you are wondering why it is like that... I do not know but one of the dependecy, `jabel` (https://github.com/Anuken/jabel), is supposed to make the Java 17 features be retro compatible with Java 8.

##### Environment
- Operating System: Windows 11
- JDK version: OpenJDK Runtime Environment Temurin-22.0.1+8 (build 22.0.1+8)
- Visual Studio Code version: 1.91.1
- Java extension version: 1.33.2024072008

##### Steps To Reproduce
1. Clone https://github.com/Anuken/MindustryJavaModTemplate
2. Add (anywhere) some simple Type Pattern cast (e.g. below)
```java
Object test = 2f;
if (test instanceof Float f) {

}
```
3. Try to build using `gradle jar` or `./grawled jar`, it should be succesful.

##### Current Result
The 'Type Pattern' are being reported as an error, even though it builds fine (and run fine in-game too, but I don't expect you to bother trying.)

##### Expected Result
The 'Type Pattern' to not be reported as an error, or at least be able to ignore the error.

##### Additional Informations
Due to the specific situation, it might not be possible to fix this, so maybe adding the option to ignore error is a possible way if it's possible for you to do.

Arthurm1:
This will take into account the `java plugin` source/target compatibility settings.
e.g. using...
```
java {
  targetCompatibility = JavaVersion.VERSION_1_8
  sourceCompatibility = JavaVersion.VERSION_1_9
}
```
instead of
```
tasks.withType(JavaCompile) {
  sourceCompatibility = JavaVersion.VERSION_1_8
  targetCompatibility = JavaVersion.VERSION_1_9
}
```

I'm not sure if this is enough to fix https://github.com/redhat-developer/vscode-java/issues/3721.  The source/target compatibility in `JVMBuildTarget` will have changed but the javac options will still contain `--release 8`.  Hopefully that will be ok for the eclipse settings.

I note that [here](https://github.com/bsideup/jabel) it says to configure the `compileJava` task with source/target and not the `java` plugin when using `Jabel`

I've also extended the `POJO` constructor copying used in `DefaultSourceSets` to the language extensions.  It's not needed for normal running but it's a pain when debugging as all the variables/watches just show the proxy classes.

### DIFF
--- a/plugin/src/main/java/com/microsoft/java/bs/gradle/plugin/JavaLanguageModelBuilder.java
+++ b/plugin/src/main/java/com/microsoft/java/bs/gradle/plugin/JavaLanguageModelBuilder.java
@@ -20,6 +20,8 @@ import java.util.stream.Stream;
 import com.microsoft.java.bs.gradle.model.GradleModuleDependency;
 import com.microsoft.java.bs.gradle.model.JavaExtension;
 import com.microsoft.java.bs.gradle.model.SupportedLanguage;
+import com.microsoft.java.bs.gradle.model.SupportedLanguages;
+import com.microsoft.java.bs.gradle.model.impl.DefaultJavaExtension;
 
 import org.gradle.api.Project;
 import org.gradle.api.file.Directory;
@@ -31,9 +33,6 @@ import org.gradle.api.tasks.compile.CompileOptions;
 import org.gradle.api.tasks.compile.JavaCompile;
 import org.gradle.plugins.ide.internal.tooling.java.DefaultInstalledJdk;
 import org.gradle.util.GradleVersion;
-
-import com.microsoft.java.bs.gradle.model.SupportedLanguages;
-import com.microsoft.java.bs.gradle.model.impl.DefaultJavaExtension;
 
 /**
  * The language model builder for Java language.
@@ -67,8 +66,10 @@ public class JavaLanguageModelBuilder extends LanguageModelBuilder {
 
       List<String> compilerArgs = getCompilerArgs(javaCompile);
       extension.setCompilerArgs(compilerArgs);
-      extension.setSourceCompatibility(getSourceCompatibility(compilerArgs));
-      extension.setTargetCompatibility(getTargetCompatibility(compilerArgs));
+      // Ignore options and get source/target compatibility directly from javaCompile.
+      extension.setSourceCompatibility(javaCompile.getSourceCompatibility());
+      extension.setTargetCompatibility(javaCompile.getTargetCompatibility());
+
       return extension;
     }
     return null;
--- a/plugin/src/test/java/com/microsoft/java/bs/gradle/plugin/GradleBuildServerPluginTest.java
+++ b/plugin/src/test/java/com/microsoft/java/bs/gradle/plugin/GradleBuildServerPluginTest.java
@@ -16,7 +16,6 @@ import java.nio.file.Paths;
 import java.util.List;
 import java.util.function.Consumer;
 import java.util.stream.Stream;
-import java.util.ArrayList;
 
 import com.microsoft.java.bs.gradle.model.GradleSourceSet;
 import com.microsoft.java.bs.gradle.model.GradleSourceSets;
@@ -57,8 +56,8 @@ class GradleBuildServerPluginTest {
         .addArguments("-Dorg.gradle.logging.level=quiet")
         .addJvmArguments("-Dbsp.gradle.supportedLanguages="
             + String.join(",", SupportedLanguages.allBspNames));
-    GradleSourceSets sourceSets = action.run();
-    return new DefaultGradleSourceSets(new ArrayList<>(sourceSets.getGradleSourceSets()));
+
+    return new DefaultGradleSourceSets(action.run());
   }
 
   private interface ConnectionConsumer {
@@ -161,7 +160,7 @@ class GradleBuildServerPluginTest {
       // JDK source/target options changed from 1.9 -> 9 in 8.0
       new GradleJreVersion("8.0", 19),
       // highest supported version
-      new GradleJreVersion("8.8", 22)
+      new GradleJreVersion("8.10.2", 22)
     ).filter(version -> version.jreVersion >= javaVersion)
      .map(GradleJreVersion::getGradleVersion);
   }
@@ -340,9 +339,10 @@ class GradleBuildServerPluginTest {
         assertFalse(args.contains("|-target|"), () -> "Available args: " + args);
         assertTrue(args.contains("|--release|9"), () -> "Available args: " + args);
         assertTrue(args.contains("|-Xlint:all"), () -> "Available args: " + args);
-        assertEquals("9", javaExtension.getSourceCompatibility(),
-            () -> "Available args: " + args);
-        assertEquals("9", javaExtension.getTargetCompatibility(),
+        String version9 = gradleVersion.compareTo(GradleVersion.version("8.0")) >= 0 ? "9" : "1.9";
+        assertEquals(version9, javaExtension.getSourceCompatibility(),
+                () -> "Available args: " + args);
+        assertEquals(version9, javaExtension.getTargetCompatibility(),
             () -> "Available args: " + args);
       }
     });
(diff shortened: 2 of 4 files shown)


jdneo:
> I've also extended the POJO constructor copying used in DefaultSourceSets to the language extensions. It's not needed for normal running but it's a pain when debugging as all the variables/watches just show the proxy classes.

There is an issue about this: https://github.com/microsoft/build-server-for-gradle/issues/175. I also did some investigation and turns out the copy constructor is not needed for running.

I'm thinking that, for debugging purpose, maybe override `toString()` is a better approach than using the copy constructor?

-----

Update:

The copy constructor is required. If we remove all the copy constructor, here:

https://github.com/microsoft/build-server-for-gradle/blob/395b97757da8d0d82bc2c68c4aca4331bea4e584/model/src/main/java/com/microsoft/java/bs/gradle/model/actions/GetSourceSetsAction.java#L174

The sourceset is a proxy wrapperred type, and it will encounter some problem 

image

plugin/src/main/java/com/microsoft/java/bs/gradle/plugin/JavaLanguageModelBuilder.java
```diff
@@ -69,6 +73,37 @@ public DefaultJavaExtension getExtensionFor(Project project, SourceSet sourceSet
       extension.setCompilerArgs(compilerArgs);
       extension.setSourceCompatibility(getSourceCompatibility(compilerArgs));
       extension.setTargetCompatibility(getTargetCompatibility(compilerArgs));
+
+      // it's possible to set source/target compatibility in different places.
+      // Java plugin setting should override javacoptions
+      JavaVersion sourceVersion = null;
+      JavaVersion targetVersion = null;
+      if (GradleVersion.current().compareTo(GradleVersion.version("7.1")) >= 0) {
```
jdneo: What the difference between getting the source/target level from the java plugin and 

```
JavaCompile .getSourceCompatibility();
JavaCompile .getTargetCompatibility();
```
?

Arthurm1: I'd like to know too 😆 

The code managing it is [here](https://github.com/gradle/gradle/blob/dd40c5b0e06bef621941984ed8ba565dbec06805/platforms/jvm/plugins-java-base/src/main/java/org/gradle/api/plugins/JavaBasePlugin.java#L327)

I think that Gradle takes any source/target compatibility the user has set for the `java` plugin and applies it to the compiler task.  And then takes the compiler task settings and applies that to javacoptions.

But any of these can be set by the user in the build file and they can clash and Gradle seems fine with that.
They can also be illegal i.e. javacoptions require `target` >= `source` but it works out because the user has also specified `--release` in the options and that takes precedence.

I guess ultimately Gradle is just feeding all these settings into javacoptions but Intellij and Eclipse want the source/target compatibility for their own compilation settings and they seem to take that from the plugin or compiler settings rather than the javacoptions.

`Jabel` also seems to be taking advantage of this and using the source/target compatibility settings even though `--release` is set.

Arthurm1:
> The copy constructor is required. If we remove all the copy constructor, here:

Shall I leave my changes to `DefaultJavaExtension`/`DefaultScalaExtension` then or do you want them gone?

jdneo:
> Shall I leave my changes to DefaultJavaExtension/DefaultScalaExtension

I think you can leave as it is. Until we find a way to get rid of the copy constructor entirely

plugin/src/main/java/com/microsoft/java/bs/gradle/plugin/JavaLanguageModelBuilder.java
jdneo: I'll file a ticket at gradle for the recommended way to get the source/target level

jdneo:
@Arthurm1

I think we should use `javaCompile.getSourceCompatibility()` and `javaCompile.getTargetCompatibility()`.

AFAIK, there are three different kind of ways to set the source/target level.

1.
```
java {
  targetCompatibility = JavaVersion.VERSION_1_8
  sourceCompatibility = JavaVersion.VERSION_1_9
}
```

2.
```
tasks.withType(JavaCompile) {
  sourceCompatibility = JavaVersion.VERSION_1_8
  targetCompatibility = JavaVersion.VERSION_1_9
}
```

3.
```
sourceCompatibility = JavaVersion.VERSION_1_8
targetCompatibility = JavaVersion.VERSION_1_9
```

The current approach cannot get the version in case 2. While using the `JavaCompile` instance can cover all three cases.

Considering `JavaCompile` is official API which no need to use reflection, I prefer to use it directly.

jdneo:
Hi @Arthurm1, would you like to update the PR?

Or do you have any concern about `javaCompile.getSourceCompatibility()` and `javaCompile.getTargetCompatibility()`

Arthurm1:
@jdneo Will look at it this weekend and probably change the PR

jdneo:
@Arthurm1 Thank you.

My thought is that we can just simply get the source/target level from JavaCompile, no need to inspecting from compile args. (Correct me if this is working for any cases)

Arthurm1:
@jdneo I've changed it so that the source/target compatibility is populated purely from the `JavaCompile` task.  `JavacOptions` are still passed back for BSP but they are ignored for compatibility settings.

This seems to reflect more how people use the Gradle settings, i.e. set source/target compatibility in `JavaCompile` for IDE usage and the `JavacOptions` for jar and release.

plugin/src/main/java/com/microsoft/java/bs/gradle/plugin/JavaLanguageModelBuilder.java
```diff
@@ -67,8 +66,10 @@ public DefaultJavaExtension getExtensionFor(Project project, SourceSet sourceSet
 
       List<String> compilerArgs = getCompilerArgs(javaCompile);
       extension.setCompilerArgs(compilerArgs);
-      extension.setSourceCompatibility(getSourceCompatibility(compilerArgs));
```
jdneo, line 70: What about `getSourceCompatibility()` && `getTargetCompatibility()`. Will you consider to remove it as well? Since I believe they are not used now.

jdneo: @Arthurm1, are you feel ok to remove these unused methods? If yes, I can append a commit to the PR and merge it.

Arthurm1: Which ones do you mean?  `JvmBuildTargetEx.getTargetCompatibility()`?  Is that no longer used?

Or do you mean `JavaLanguageModelBuilder.getTargetCompatibility(List<String> compilerArgs)` should be removed?  ~That's a copy of what Gradle does [here](https://github.com/gradle/gradle/blob/dbd2232425b245f077cacb58b9c5acbc1690511e/platforms/jvm/language-java/src/main/java/org/gradle/api/internal/tasks/compile/JavaCompilerArgumentsBuilder.java#L143).  It looks like we should probably add more option handling from the Gradle code.~ Wrong link.
That's a copy of what Gradle does [here](https://github.com/gradle/gradle/blob/d4cc98c931d8e0972d2fc4ad03f7c5a186c2d37e/platforms/jvm/language-java/src/main/java/org/gradle/api/tasks/compile/JavaCompile.java#L295)

jdneo: Sorry, my bad. I was not aware that they are used when parsing the compiler args.

jdneo (approved):

jdneo:
Thank you @Arthurm1

PR #207, opened 2025-09-24
feat: support gradle 9.1

wenytang-ms:
https://docs.gradle.org/current/userguide/compatibility.html


chagong (approved):

server/src/test/java/com/microsoft/java/bs/core/internal/gradle/UtilsTest.java
```diff
@@ -11,6 +11,7 @@
 import java.io.File;
 import java.nio.file.Paths;
 
+import org.eclipse.xtend.lib.DataProcessor.Util;
```
Arthurm1: Is this needed?

wenytang-ms: updated

chagong (approved):

